PR #62, opened 2018-04-26
Fix for ory/oathkeeper/issues/61

andre-perry-deidle:


arekkas:
This PR resolves issues with the current master branch of Hydra, right?

andre-perry-deidle:
Yes, but it's not complete. I think it lacks tests.

arekkas:
Thank you for your contribution. As discussed in the chat, this will be solved by #60

PR #102, opened 2018-08-14
proxy: add cookies credentials issuer

zikes:
This Credentials Issuer is very similar to the [Headers](https://github.com/ory/oathkeeper/pull/100) one, making it very easy to put together initially, but it was complicated a bit by all cookies being stored in a single HTTP header. Unfortunately the `AddCookie` method of `http.Header` will just keep appending cookies of the same name over and over and upstream has to just decide for itself which one to pay attention to, so I had to cache incoming request's cookies, remove them all, then re-add any that aren't being handled by the CI rule. This should prevent upstream from receiving spoofed cookies. Thus far testing this has proved stable.

Cookie values are determined in the same way as headers, with `text/template` and a `print` FuncMap function:

```json
{
    "id": "some-id",
    "upstream": {"url": "http://my-backend-service"},
    "match": { },
    "authenticators": [ ],
    "authorizer": { },
    "credentials_issuer": {
        "handler": "cookies",
        "config": {
            "cookies": {
                "user": "{{ print .Subject }}",
                "audience": "{{ print .Extra.aud }}",
                "issuer": "{{ print .Extra.iss }}",
                "arbitrary": "{{ print .Extra.some.deeply.nested.value }}"
            }
        }
    }
}
```

### DIFF
--- a/proxy/credentials_issuer_cookies.go
+++ b/proxy/credentials_issuer_cookies.go
@@ -1,0 +1,99 @@
+package proxy
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"text/template"
+
+	"github.com/ory/oathkeeper/rule"
+	"github.com/pkg/errors"
+)
+
+type CredentialsCookiesConfig struct {
+	Cookies map[string]string `json:"cookies"`
+}
+
+type CredentialsCookies struct {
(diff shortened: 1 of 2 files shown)


zikes:
Docs PR at https://github.com/ory/docs/pull/44

PR #116, opened 2018-08-31
cmd: Enables TLS option on serve api

fredbi:
Signed-off-by: Frederic BIDON <frederic@oneconcern.com>

NOTE: now hydra, keto and oathkeeper follow a homogeneous set of options regarding TLS setup

Nice to have next: factorize between repos


aeneasr:
Thank you!

PR #123, opened 2018-10-10
Extend available transports for tokens: authorizers "jwt" and "oauth2_introspection"

fredbi:
now support tokens coming under other forms than the "Authorization Bearer" header.

* RFC6750 bearer tokens: as query param (?access_token=...) or form param
* token in cookie (atm, support is provided for base64 encoded JSON values for cookies)

Signed-off-by: Frederic BIDON <frederic@oneconcern.com>

### DIFF
--- a/helper/bearer.go
+++ b/helper/bearer.go
@@ -21,15 +21,116 @@
 package helper
 
 import (
+	"encoding/base64"
+	"encoding/json"
+	"mime"
 	"net/http"
 	"strings"
+
+	"github.com/pkg/errors"
 )
 
-func BearerTokenFromRequest(r *http.Request) string {
+const accessToken = "access_token"
+
+// ErrInvalidCookieFormat indicates a malformed configuration
+var ErrInvalidCookieFormat = errors.New("invalid cookie format specified. Supported values are: 'rawstd-base64-json' (default), 'rawurl-base64-json'")
+
+// BearerTokenFromRequest retrieves an access token from request according
+// to RFC6750
+func BearerTokenFromRequest(r *http.Request) (token string) {
 	auth := r.Header.Get("Authorization")
 	split := strings.SplitN(auth, " ", 2)
-	if len(split) != 2 || !strings.EqualFold(split[0], "bearer") {
-		return ""
+	if len(split) == 2 && strings.EqualFold(split[0], "bearer") {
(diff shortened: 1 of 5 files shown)


aeneasr:
Hi, thank you for this contribution! Would it be possible to make two PRs (or at least two commits) out of the two features? This will make review easier, the git history better, and will address another issue:

What's the use case for cookies? I understand that cookies are a place to store credentials in but we're typically seeing cookies as something that is issued by a trusted system (encrypted cookies, signed cookies, database reference) and having base64 encoded JSON here seems very bespoke/custom/non-standard to me. Is there any standard or spec this can be referenced to? Why is it base64? Why not a JWT that can be signed/verified? Or some other encryption/signing mechanism? There are so many ways cookies work and I am not aware of a standard that addresses this in the realm of OAuth2, I'm not sure this should be something for us to provide. In general, I would also avoid unsigned cookies as you can't trust the client side. I'm aware that oauth2 access tokens are a bit special here because we have a third authority, but you never know where this project will be headed in the future and that might be a complication.

aeneasr:
Ok I slept over this. I think cookies are a good idea to support. It makes sense to have a way to validate credentials coming from cookies. If your server sets the cookie after oauth2 it greatly improves developer ux. There still the issue with the format though. Is there maybe some standard we can lean on? I want to avoid us re-inventing authZ patterns.

If you fragment this PR into two (query param, cookie) I can definitely merge the first one right away.

fredbi:
yes format is a pain. besides JWT, I am looking into formats produced by express.sessions and gorilla/securecookies to cover popular toolkits

aeneasr:
Sorry for being silent on this, it dropped of my radar. Could you please split this PR into two:

1. RFC6750 bearer tokens: as query param (?access_token=...) or form param
2. token in cookie (atm, support is provided for base64 encoded JSON values for cookies)

I really want to merge the first PR but the second has to be discussed further. I disagree with the approach of catering to every framework out there, it will be a maintainer's and adopter's nightmare when those things change and it will prompt people to just add PR over PR to introduce Spring, PHP, and all of those things we don't want to deal with.

fredbi:
I am afraid there is no practical solution to the cookie problem within these constraints. Giving up.

PR #139, opened 2018-11-29
Ignore query parameters when matching url in rules.

stszap:
It seems that oathkeeper compares urls without stripping GET parameters. For example, a rule with
```
    "match": {
        "url": "http://test.com/test/",
        "methods": ["GET"]
    }
```
matches `http://test.com/test/` but doesn't match `http://test.com/test/?param=value`. [Comments](https://github.com/ory/oathkeeper/blob/master/rule/rule.go#L43) say:

> it compares the full request URL (e.g. https://mydomain.com/api/resource) **without query parameters** of the incoming request

So I guess it's a bug? According to [the docs](https://golang.org/pkg/net/url/#URL.String) .String() method uses `scheme://userinfo@host/path?query#fragment` pattern so I tried to fix it by simple Sprintf() and added a test.

## Checklist

- [x] I have read the [contributing guidelines](CONTRIBUTING.md)
vulnerability, I confirm that I got green light (please contact [hi@ory.sh](mailto:hi@ory.sh)) from the maintainers to push the changes.
- [x] I signed the [Developer's Certificate of Origin](https://github.com/ory/keto/blob/master/CONTRIBUTING.md#developers-certificate-of-origin)
by signing my commit(s). You can amend your signature to the most recent commit by using `git commit --amend -s`. If you
amend the commit, you might need to force push using `git push --force HEAD:<branch>`. Please be very careful when using
force push.

### DIFF
--- a/rule/rule.go
+++ b/rule/rule.go
@@ -22,6 +22,7 @@ package rule
 
 import (
 	"encoding/json"
+	"fmt"
 	"hash/crc32"
 	"net/url"
 	"regexp"
@@ -124,7 +125,7 @@ func (r *Rule) IsMatching(method string, u *url.URL) error {
 		return errors.WithStack(err)
 	}
 
-	if !c.MatchString(u.String()) {
+	if !c.MatchString(fmt.Sprintf("%s://%s%s", u.Scheme, u.Host, u.Path)) {
 		return errors.Errorf("rule %s does not match URL %s", r.ID, u)
 	}
 
--- a/rule/rule_test.go
+++ b/rule/rule_test.go
@@ -43,5 +43,6 @@ func TestRule(t *testing.T) {
 	}
 
 	assert.NoError(t, r.IsMatching("DELETE", mustParse(t, "https://localhost/users/1234")))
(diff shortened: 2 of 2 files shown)


aeneasr:
Nice catch. Maybe we should make this configurable? Just merging this could leave people vulnerable that think it's matching query params? What do you think?

stszap:
Well, this behavior was only mentioned in the comments and the docs are not clear about it. I can try to add "ignore_query_params" boolean option to "match" section if security of current users is a concern.

stszap:
I think that verifying individual parameters may me too much. It couples an application with oathkeeper and may require  pretty complex syntax/options. There can be users who only want to validate certain parameters and just ignore the rest, and other users who want to validate every parameter and deny everything else. Some of params can be arrays and some can be strings. I don't have any concrete arguments agains it but my gut feeling is that get parameters should be left for an application to deal with and there should be a way for oathkeeper to ignore them.

aeneasr:
Yeah, that makes sense! So let's keep the PR as is and have it ignore query params completely for now. Thank you!

PR #211, opened 2019-06-17
Add cookie session authenticator

alexdavid:
## Related issue

@aeneasr as discussed this is an initial implementation of what the cookie session authenticator could look like. If you like the API I can go in and add some documentation around it, otherwise please let me know if you have a better idea of how you would like this to work.

## Proposed changes
The authenticator assumes there's a RESTful session service that responds to `GET /sesions/{sessionId}` with the json structure `{ subject: string, extra: map[string]interface{} }`. This authenticator reads a session cookie from the incoming request. Then if the passed sessionId returns a 200, it populates the subject/extra fields for the pipeline. If the cookie does not exist it returns `ErrAuthenticatorNotResponsible`, otherwise authentication fails.

## Checklist
- [X] I have read the [contributing guidelines](CONTRIBUTING.md)
- [X] I have read the [security policy](SECURITY.md)
      the changes.
- [X] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/docs/config.yaml
+++ b/docs/config.yaml
@@ -255,6 +255,20 @@ authenticators:
     # Sets the anonymous username. Defaults to "anonymous". Common names include "guest", "anon", "anonymous", "unknown".
     subject: anonymous
 
+  # Configures the cookie session authenticator
+  cookie_session:
+    # Set enabled to true if the authenticator should be enabled and false to disable the authenticator. Defaults to false.
+    enabled: true
+
+    # Sets the origin to proxy requests to. If the response is a 200 with body `{ "subject": "...", "extra": {} }`
+    # The request will pass the subject through successfully, otherwise it will be marked as unauthorized
+    check_session_url: https://session-store-host
+
+    # Sets a list of possible cookies to look for on incoming requests, and will fallthrough to the next authenticator if
+    # none of the passed cookies are set on the request
+    only:
+      - sessionid
+
   # Configures the jwt authenticator
   jwt:
     # Set enabled to true if the authenticator should be enabled and false to disable the authenticator. Defaults to false.
--- a/driver/configuration/provider_viper_authn.go
+++ b/driver/configuration/provider_viper_authn.go
@@ -18,6 +18,11 @@ const (
 	// noop
 	ViperKeyAuthenticatorNoopIsEnabled = "authenticators.noop.enabled"
 
+	// cookie session
+	ViperKeyAuthenticatorCookieSessionIsEnabled       = "authenticators.cookie_session.enabled"
+	ViperKeyAuthenticatorCookieSessionCheckSessionURL = "authenticators.cookie_session.check_session_url"
+	ViperKeyAuthenticatorCookieSessionOnly            = "authenticators.cookie_session.only"
+
 	// jwt
 	ViperKeyAuthenticatorJWTIsEnabled     = "authenticators.jwt.enabled"
 	ViperKeyAuthenticatorJWTJWKSURIs      = "authenticators.jwt.jwks_urls"
@@ -51,6 +56,18 @@ func (v *ViperProvider) AuthenticatorAnonymousIdentifier() string {
 func (v *ViperProvider) AuthenticatorNoopIsEnabled() bool {
 	return viperx.GetBool(v.l, ViperKeyAuthenticatorNoopIsEnabled, false)
 
+}
+
+func (v *ViperProvider) AuthenticatorCookieSessionIsEnabled() bool {
+	return viperx.GetBool(v.l, ViperKeyAuthenticatorCookieSessionIsEnabled, false)
+}
+
+func (v *ViperProvider) AuthenticatorCookieSessionCheckSessionURL() string {
+	return viperx.GetString(v.l, ViperKeyAuthenticatorCookieSessionCheckSessionURL, "")
+}
+
+func (v *ViperProvider) AuthenticatorCookieSessionOnly() []string {
+	return viperx.GetStringSlice(v.l, ViperKeyAuthenticatorCookieSessionOnly, []string{})
 }
 
 func (v *ViperProvider) AuthenticatorJWTIsEnabled() bool {
--- a/pipeline/authn/authenticator_cookie_session.go
+++ b/pipeline/authn/authenticator_cookie_session.go
@@ -1,0 +1,136 @@
+package authn
+
+import (
+	"bytes"
+	"encoding/json"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+
+	"github.com/pkg/errors"
+
+	"github.com/ory/oathkeeper/driver/configuration"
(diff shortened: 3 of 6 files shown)


alexdavid:
@kevgo

aeneasr:
Awesome, thank you! I didn't have time to review it yet but it's on my list for this week.

pipeline/authn/authenticator_cookie_session.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	return &AuthenticationSession{
+		Subject: subject,
+		Extra:   extra,
+	}, nil
+}
+
+func getSession(cf AuthenticatorCookieSessionConfiguration, sessionId string) (string, map[string]interface{}, error) {
+	resp, err := http.Get(cf.SessionStoreOrigin + "/sessions/" + sessionId)
```
aeneasr: This makes some assumptions about the upstream store. While I think that we can't avoid it (e.g. in session) I think it should be avoided here. What do you think about a request that simply forwards the incoming request + headers to the upstream? Check out how ambassador is doing it: https://www.getambassador.io/reference/services/auth-service/#the-external-authentication-service

That would also remove the need for `cf.CookieName`

pipeline/authn/authenticator_cookie_session.go
```diff
@@ -0,0 +1,97 @@
...
+	if err != nil {
+		return "", nil, err
+	}
+
+	if resp.StatusCode != 200 {
+		return "", nil, errors.New("Session not found")
```
aeneasr: This should probably be something like ErrUnauthorized

pipeline/authn/authenticator_cookie_session.go
alexdavid: I like that a lot better. Do you think we should send the full request through, or normalize it into a GET or HEAD request? (I'm thinking if there's a post request with a large request body like a file upload the session store probably doesn't need all that data sent to it).

alexdavid: Ah I see ambassador just forwards the method + headers and has a 0 length body. That sounds good to me. Will update shortly

alexdavid: One issue with removing the need for `cf.CookieName` is if the user wants a `cookie-session` authn and an `anonymous` authn we can't return `ErrAuthenticatorNotResponsible` if we don't know the cookie name to switch on. Would you prefer this happen through return status code of the session validator service or by making `CookieName` an optional option that allows fallthrough if set? I have my concerns about return status code because really the most appropriate return status code would probably be `400`, but I think it's important that any `4xx-5xx` value means invalid credentials.

aeneasr: > Ah I see ambassador just forwards the method + headers and has a 0 length body. That sounds good to me. Will update shortly

Exactly :)

> One issue with removing the need for cf.CookieName is if the user wants a cookie-session authn and an anonymous authn we can't return ErrAuthenticatorNotResponsible if we don't know the cookie name to switch on. 

How about an optional filter: `only: foo,bar,baz` -  or alternatively `without: foo,bar`

alexdavid:
@aeneasr comments addressed, let me know what you think!

docs/config.yaml
```diff
@@ -255,6 +255,20 @@ authenticators:
...
     subject: anonymous
 
+  # Configures the cookie session authenticator
+  cookie_session:
+    # Set enabled to true if the authenticator should be enabled and false to disable the authenticator. Defaults to false.
+    enabled: true
+
+    # Sets the origin to proxy requests to. If the session store returns a 200 with body `{ "subject": "...", "extra": {} }`
+    # The request will pass the subject through successfully, otherwise it will be marked as unauthorized
+    session_store_origin: https://origin-of-session-store
```
aeneasr: To stay consistent with other configs, how about `session_store_url`? Or maybe `check_session_url`?

driver/configuration/provider_viper_authn.go
```diff
@@ -53,6 +58,18 @@ func (v *ViperProvider) AuthenticatorNoopIsEnabled() bool {
 
 }
 
+func (v *ViperProvider) AuthenticatorCookieSessionIsEnabled() bool {
+	return viperx.GetBool(v.l, ViperKeyAuthenticatorCookieSessionIsEnabled, false)
+}
+
+func (v *ViperProvider) AuthenticatorCookieSessionSessionStoreOrigin() string {
+	return viperx.GetString(v.l, ViperKeyAuthenticatorCookieSessionSessionStoreOrigin, "", "AUTHENTICATOR_COOKIE_SEESSION_SESSION_STORE_ORIGIN")
```
aeneasr: You can omit `AUTHENTICATOR_COOKIE_SEESSION_SESSION_STORE_ORIGIN`. The last parameters are used for backwards compatible changes (e.g. when the config key changes you can add it to that list and it will look for the values and throw potentially a warning).

driver/configuration/provider_viper_authn.go
```diff
@@ -53,6 +58,18 @@ func (v *ViperProvider) AuthenticatorNoopIsEnabled() bool {
...
+}
+
+func (v *ViperProvider) AuthenticatorCookieSessionOnly() []string {
+	return viperx.GetStringSlice(v.l, ViperKeyAuthenticatorCookieSessionOnly, []string{}, "AUTHENTICATOR_COOKIE_SESSION_ONLY")
```
aeneasr: Same here

pipeline/authn/authenticator_cookie_session.go
```diff
@@ -0,0 +1,128 @@
...
+		c: c,
+	}
+}
+
+func (a *AuthenticatorCookieSession) GetID() string {
+	return "cookie_session"
+}
+
+func (a *AuthenticatorCookieSession) Validate() error {
+	if !a.c.AuthenticatorCookieSessionIsEnabled() {
```
aeneasr: We should also check if the check session url is set!

pipeline/authn/authenticator_cookie_session.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/ory/oathkeeper/helper"
+	"github.com/ory/oathkeeper/pipeline"
+)
+
+type AuthenticatorCookieSessionFilter struct {
+}
+
+type AuthenticatorCookieSessionConfiguration struct {
+	Only               []string `json:"only"`
+	SessionStoreOrigin string   `json:"session_store_origin"`
```
aeneasr: If the value is renamed in the config, it should be changed here too

docs/config.yaml
alexdavid: Sounds good, while this is technically an origin since it's protocol + host but no path, I guess the other "URL"s are too. Consistency is probably better here. I like `check_session_url`!

aeneasr:
Awesome, thank you!

PR #235, opened 2019-08-08
Add sprig template library

hypnoglow:
## Proposed changes

This commit adds [sprig](https://github.com/Masterminds/sprig) template library functions to the templates used in header/cookie mutator.

Go stdlib templates are very limited in terms of features. For example, I have a use case where I have an array of strings in session data, and I want to use mutator to join them into the single header string. Currently it is kind of impossible to do properly. This template almost does this, but there is no possibility on removing trailing comma:

```
{{- $claims := (index .Extra "example.com/some-claims") -}}
{{- range $claim := $claims }}
{{- $claim -}},
{{- end -}}
```

With Extra set to `map[string]interface{}{"example.com/some-claims":[]string{"Foo","Bar"}` it compiles into `Foo,Bar,`.

With sprig library we could replace the whole template with simple `{{- (index .Extra "example.com/some-claims") | join "," -}}` and it will produce correct output `Foo,Bar`.

Sprig can also handle many other advanced use cases.

## Checklist


aeneasr:
This is a great idea!

Issue #228, opened 2019-07-25
Mutator issuing  JWT with custom claims


piotrmsc:
**Is your feature request related to a problem? Please describe.**

Feature request :)  open for contribution ;-) 

**Describe the solution you'd like**
Oathkeeper has id token mutator which issues JWT and with id token calls the upstream. What would be nice to have is a mutator that issues JWT but allows to add custom claims to that JWT, based on input provided from different service.

Additional claims could be used in the mutator by for example fetching  information (claims) from a different place (can be a rest/graphql/grpc service) which provides information for the given request. 

This could be used in scenarios when you want to call the upstream in the oathkeeper with some additional context such as tenant using claims in in JWT and perhaps even headers.
**Describe alternatives you've considered**

**Additional context**

aeneasr:
Any ideas on the API / configuration side?

piotrmsc:
If we think of new mutator then I think the current API (the interface) does not have to be changed. 
New mutator would just fetch additional claim from the given place (http service,grps,graphql or even some kind of storage) and add it in the `Mutate` method of https://github.com/ory/oathkeeper/blob/master/pipeline/mutate/mutator.go#L39.

If we think of extending current JWT mutator then I think we have the same situation as with new one.

I think that then access rule should have configuration extended to point from where fetch additional claims perhaps on the mutator level. What do you think?

aeneasr:
For this to work we decided to:

- Change mutator from single to multiple (similar to authenticator)
- Introduce an RPC Mutator (or "enhancer" - name TBD)
- Update the ID Token Mutator to make it possible to define custom claims easily

## Mutator changes

We would allow our mutators to handle multiple steps. Mutators are called from top to bottom. If a mutator fails with an error, we send a 403 or 500 error code and stop propagation.
Mutators can mutate the state of [AuthenticationSession](https://github.com/ory/oathkeeper/blob/master/pipeline/authn/authenticator.go#L26-L29).

```
  {
    "id": "test-rule-2",
    "upstream": { ... },
    "match": { ... },
    "authenticators": [ ... ],
    "authorizer": [ ... ],
-    "mutator": {
-      "handler": "id_token"
-    }
+   "mutators": [
+     {
+        "handler": "enhancer",
+      },
+     {
+        "handler": "id_token",
+      }
+   ]
+},
```

### RPC Mutator

*We still need a good name here.*

[...]

PR #243, opened 2019-08-14
Extend ID_TOKEN mutator with custom claims

jakkab:
## Related issue

#228 

## Proposed changes

- Mutator ID_TOKEN now supports go templates
- Users can define custom claims in rule configuration

## Checklist

- [x] I have documented my changes in the [developer guide](https://github.com/ory/docs) (if appropriate)

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -44,6 +44,33 @@ before finalizing the upgrade process.
 1. ORY Oathkeeper now supports multiple mutators. Mutations are performed in the provided order and must all succeed in order for the HTTP request to be forwarded.
 2. The `mutator` property was renamed to `mutators` to reflect its true nature (see previous item).
 
+#### id_token mutator
+
+The id_token mutator now allows to specify custom claims via the `claims` field of the mutator's `config` field. The keys represent names of claims and the values are strings which will be parsed by the Go [text/template](https://golang.org/pkg/text/template/) package for value substitution, receiving the `AuthenticationSession` struct. You must now specify
+the audience of the ID token directly via the `aud` claim. If you use `id_token` mutators with defined audience, please update them as follows:
+
+deprecated config:
+```json
+{
+  "handler": "id_token",
+  "config": {
+    "aud": ["https://my-backend-service/some/endpoint"]
+  }
+}
+```
+
+new config:
+```json
+{
+  "handler": "id_token",
+  "config": {
+    "claims": {
+      "aud": ["https://my-backend-service/some/endpoint"]
+    }
+  }
+}
+```
+
 ### Access Rule Changes
 
 As already noted, the `mutator` property was renamed to `mutators` and now represents a list of mutation handlers. If you have
--- a/pipeline/mutate/mutator_id_token.go
+++ b/pipeline/mutate/mutator_id_token.go
@@ -24,14 +24,17 @@ import (
 	"bytes"
 	"encoding/json"
 	"net/http"
+	"text/template"
 	"time"
+
+	"github.com/ory/x/jsonx"
+
+	"github.com/dgrijalva/jwt-go"
 
 	"github.com/ory/oathkeeper/credentials"
 	"github.com/ory/oathkeeper/driver/configuration"
 	"github.com/ory/oathkeeper/pipeline"
 	"github.com/ory/oathkeeper/pipeline/authn"
-
-	"github.com/dgrijalva/jwt-go"
 	"github.com/pborman/uuid"
 	"github.com/pkg/errors"
 )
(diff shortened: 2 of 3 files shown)


pipeline/mutate/mutator_id_token.go
```diff
@@ -41,23 +44,29 @@ type MutatorIDTokenRegistry interface {
 }
 
 type CredentialsIDTokenConfig struct {
-	Audience []string `json:"aud"`
+	Audience []string          `json:"aud"`
```
aeneasr: I guess the question is if we want to keep this or not. As now we can use the claims portion instead.

jakkab: True. Do you want me to delete it?

aeneasr: Yeah, I think that makes sense - but we have to document that in the upgrade guide .)

jakkab: Sure! I'm on it!

UPGRADE.md
```diff
@@ -44,6 +44,33 @@ before finalizing the upgrade process.
...
+}
+```
+
+new config:
+```json
+{
+  "handler": "id_token",
+  "config": {
+    "claims": {
+      "aud": ["https://my-backend-service/some/endpoint"]
```
aeneasr: I think this makes it particularly hard to make use of these claims. I think the escape workaround ( https://github.com/ory/oathkeeper/pull/243/files#diff-5ab69f721ec10aff3758f597a3aa0de6R84 ) is also related to this. I had another idea, what do you think about using a raw json object in claims? So then you could do:

```json
{
  "aud": {{ .Whatever }}
}
```

or for a full example (**note: this needs sprig or "quote" will not work**):

```go
package main

import (
	"os"
	"text/template"
)

var card2 = map[string]string{"a": "whatever", "b": "bazbar"}

func check(err error) {
	if err != nil {
		panic(err)
	}
}

func main() {
	// Use {{define}} to prevent write multiple times.
	t := template.Must(template.New("t2").Parse(`{
	"aud": [{{ .whatever | quote }}, {{ .b | quote }}]
}`))
	check(t.Execute(os.Stdout, card2))
}
```

So basically `claims` would no longer be a `map[string]string` but instead just raw bytes. I think it makes sense to send the compiled template using `debug` log level if the compilation of the template fails.

This should then also help resolve and/or debug the e2e test failure.

What do you think?

aeneasr: The above example would compile to:

```json
{
  "aud": ["whatever", "bazbar"]
}
```

jakkab: Using raw bytes IS the problem, I'm afraid. Parse func accepts a template like
```
`"{{- (index .Extra \"example.com/some-claims\") | join \",\" -}}"`
```
only if it has been decoded to string. Decoding claims to interface{} or even a structure like
```
type CredentialsIDTokenConfig struct {
	Claims json.RawMessage `json:"claims"`
}
```
preserves escape characters and causes Parse() to fail due to unexpected operands.

jakkab: see example: https://play.golang.org/p/m_9E7DlvHir

jakkab: Using raw bytes IS the problem, I'm afraid. Parse() refuses to digest something like
```
`{"custom-claim": "{{- (index .Extra \"example.com/some-claims\") | join \",\" -}}"}`
```
due to unexpected "\\" in operand. That's reasonable, since the sequence contains escape characters. If unmarshalled to a string, however, the
```
"{{- (index .Extra \"example.com/some-claims\") | join \",\" -}}"
```
expression can be parsed (escaped characters have been removed,) but bear in mind we allow arbitrary data structures, not just strings.

jakkab: Obviously, we could drop them backslashes from the config itself:
```
`{"claims":{"custom-claim": "{{- (index .Extra "example.com/some-claims") | join "," -}}", "aud": ["foo", "bar"]}}`
```
but they are required to unmarshal the object properly, so decoding simply won't work. 
We could even skip decoding by passing the entire "config" object (with no backslashes, as above) to Parse(). Template will be executed just fine, but this approach involves config validation (we allow only one key in the top-level map, i.e. "claims") and requires some further type assertions.

aeneasr: This works fine though?

```go
package main

import (
	"os"
	"text/template"
)

type card struct {
	Extra map[string]interface{}
}

const jtemplate = `{
	"custom-claim": "{{- (index .Extra "example.com/some-claims") -}}"
}`

func main() {
	t := template.Must(template.New("t2").Parse(jtemplate))
	_ = t.Execute(os.Stdout, &card{Extra: map[string]interface{}{"example.com/some-claims": []string{"foo", "bar"}}})
}
```

Prints:

```
{
	"custom-claim": "[foo bar]"
}
```

Am I missing something?

jakkab: The example works fine, cause no decoding is involved. If you try to unmarshal a complete, unescaped "claims" object like this
```
`{"claims":{"custom-claim": "{{- (index .Extra \"example.com/some-claims\") | join \",\" -}}", "aud": ["foo", "bar"]}}`
```
the decoder simply goes nuts, because it expects a comma or another suitable conclusion to a key:value pair ("custom-claim": **"{{- (index .Extra \"**), but gets "example" instead.
See https://play.golang.org/p/jupXW2QTk8y.

The alternative involves parsing the config directly. In fact, it is going to work, but leaves us with a raw message (executed template) that still requires unmarshalling:
https://play.golang.org/p/v01iH3HLdig.

Is the second approach acceptable?

aeneasr: The first example panics because jtemplate is the Go template - the json en/decoder can only work raw json, not go templates. Therefore it is necessary to compile the template first and then de/encode it to json! 

The second approach is more than acceptable, I think it's the only solution to this problem :)

aeneasr:
If you need help with the e2e tests lmk

pipeline/mutate/mutator_id_token.go
```diff
@@ -40,47 +42,61 @@ type MutatorIDTokenRegistry interface {
...
+		if err != nil {
+			return errors.Wrapf(err, `error parsing claims template in rule "%s"`, rl.GetID())
 		}
 	}
 
-	if len(cc.Audience) > 0 {
-		claims["aud"] = cc.Audience
+	b := bytes.Buffer{}
+	err = tmpl.Execute(&b, session)
+	if err != nil {
```
aeneasr:
```suggestion
if err := tmpl.Execute(&b, session); err != nil
```

pipeline/mutate/mutator_id_token.go
```diff
@@ -40,47 +42,61 @@ type MutatorIDTokenRegistry interface {
...
+		tmpl, err = a.t.New(templateID).Parse(string(config))
+		if err != nil {
+			return errors.Wrapf(err, `error parsing claims template in rule "%s"`, rl.GetID())
 		}
 	}
 
-	if len(cc.Audience) > 0 {
-		claims["aud"] = cc.Audience
+	b := bytes.Buffer{}
+	err = tmpl.Execute(&b, session)
```
aeneasr:
```suggestion
```

pipeline/mutate/mutator_id_token.go
```diff
@@ -40,47 +42,61 @@ type MutatorIDTokenRegistry interface {
...
+	if err != nil {
+		return errors.Wrapf(err, `error executing claims template in rule "%s"`, rl.GetID())
+	}
+
+	var cc CredentialsIDTokenConfig
+	d := json.NewDecoder(bytes.NewBuffer(b.Bytes()))
+	d.DisallowUnknownFields()
+	if err = d.Decode(&cc); err != nil {
```
aeneasr: This can be a oneliner: `github.com/ory/x/jsonx.NewStrictDecoder()`

```suggestion
	if err := jsonx.NewStrictDecoder(bytes.NewBuffer(b.Bytes())).Decode(&cc); err != nil {
```

pipeline/mutate/mutator_id_token.go
```diff
@@ -40,47 +42,61 @@ type MutatorIDTokenRegistry interface {
...
-	if len(cc.Audience) > 0 {
-		claims["aud"] = cc.Audience
+	b := bytes.Buffer{}
+	err = tmpl.Execute(&b, session)
+	if err != nil {
+		return errors.Wrapf(err, `error executing claims template in rule "%s"`, rl.GetID())
+	}
+
+	var cc CredentialsIDTokenConfig
+	d := json.NewDecoder(bytes.NewBuffer(b.Bytes()))
```
aeneasr:
```suggestion
```

pipeline/mutate/mutator_id_token.go
```diff
@@ -40,47 +42,61 @@ type MutatorIDTokenRegistry interface {
...
+	d.DisallowUnknownFields()
```
aeneasr:
```suggestion
```

pipeline/mutate/mutator_id_token.go
```diff
@@ -40,47 +42,61 @@ type MutatorIDTokenRegistry interface {
...
+	var tmpl *template.Template
+	var err error
 
-	now := time.Now().UTC()
-	claims := jwt.MapClaims{}
-	if session.Extra != nil {
-		for k, v := range session.Extra {
-			claims[k] = v
+	templateID := rl.GetID()
+	tmpl = a.t.Lookup(templateID)
```
aeneasr: It makes sense to define the variable where it's actually being used/defined - this prevents potential nil panics in future code changes.

```suggestion
	tmpl := a.t.Lookup(templateID)
```

pipeline/mutate/mutator_id_token.go
```diff
@@ -40,47 +42,61 @@ type MutatorIDTokenRegistry interface {
...
 		config = []byte("{}")
 	}
 
-	var cc CredentialsIDTokenConfig
-	d := json.NewDecoder(bytes.NewBuffer(config))
-	d.DisallowUnknownFields()
-	if err := d.Decode(&cc); err != nil {
-		return errors.WithStack(err)
-	}
+	var tmpl *template.Template
```
aeneasr:
```suggestion
```

pipeline/mutate/mutator_id_token.go
```diff
@@ -40,47 +42,61 @@ type MutatorIDTokenRegistry interface {
...
 
-	now := time.Now().UTC()
-	claims := jwt.MapClaims{}
-	if session.Extra != nil {
-		for k, v := range session.Extra {
-			claims[k] = v
+	templateID := rl.GetID()
+	tmpl = a.t.Lookup(templateID)
+	if tmpl == nil {
+		tmpl, err = a.t.New(templateID).Parse(string(config))
```
aeneasr: This should be `Prase(string(CredentialsIDTokenConfig.Claims)`

jakkab: Wait, are you sure? I thought we execute the template first to avoid those parsing errors!

aeneasr:
Thank you so much for your work - I've made some changes to this PR and merged it with #246 :)

PR #246, opened 2019-08-20
ID Token Custom Claims

aeneasr:
/cc @jakkab


aeneasr:
You probably need to filter out *.md files - they got reformatted by accident

jakkab:
see #228

PR #258, opened 2019-09-19
Homogenize configuration management

aeneasr:
To do:

- [x] `mutator.hydrator.api.retry.delayInMilliseconds` => `mutator.hydrator.config.api.retry.delay_in_milliseconds`
- [x] `mutator.hydrator.api.retry.number` => `mutator.hydrator.api.retry.number_of_retries`
- [x] Replace json schema reference `/fix-255/` with `/master/`
- [x] Revert https://github.com/ory/oathkeeper/pull/258/files#diff-9f6f74101514723a98e882f06779ada8R53
- [x] Revert https://github.com/ory/oathkeeper/pull/258/files#diff-4c51d95b26718af4a2ac3bec33e54e49L132-L135

mutators header and cookie now need to be configured in the main config like this:

```
  header:
    enabled: true
    config:
      headers:
        foo: bar

  # Configures the cookie mutator
  cookie:
    # Set enabled to true if the mutator should be enabled and false to disable the mutator. Defaults to false.
    enabled: true
    config:
      cookies:
        foo: bar
```

keto like this:

```

  # Configures the keto_engine_acp_ory authorizer
  keto_engine_acp_ory:
    # Set enabled to true if the authorizer should be enabled and false to disable the authorizer. Defaults to false.
    enabled: true

    config:
      # REQUIRED IF ENABLED - The base URL of ORY Keto, typically something like http(s)://<host>[:<port>]/
      base_url: http://my-keto/
      required_action: unknown
      required_resource: unknown
```


cmd/server/server.go
```diff
@@ -129,10 +129,7 @@ func cert(daemon string, logger logrus.FieldLogger) []tls.Certificate {
 func clusterID(c configuration.Provider) string {
 	var id bytes.Buffer
 	if err := json.NewEncoder(&id).Encode(viper.AllSettings()); err != nil {
-		for _, repo := range append(
-			append(c.AccessRuleRepositories(), *c.MutatorIDTokenJWKSURL(), *c.MutatorIDTokenIssuerURL()),
-			c.AuthenticatorJWTJWKSURIs()...,
-		) {
+		for _, repo := range c.AccessRuleRepositories() {
```
aeneasr: revert

PR #265, opened 2019-09-24
Traefik auth forward

mvanderlee:
Behaves identical to /decisions except that it gets the url and method from the headers.
This allows integration with Traefik's ForwardAuth middleware.

## Related issue

#263 

## Proposed changes

Add integration with Traefik by adding a new `/auth_forward` endpoint that behaves identical to `/decisions` except that it gets the url and method from the following headers:

- `X-Forwarded-Uri`
- `X-Forwarded-Method`

## Checklist

- [ ] I have read the [security policy](../security/policy)
      the changes.

## Further comments

I've copied the `decision.go` file, renamed where appropriate, and changed the rule matcher input. Found references to `DecisionHandler` and added the new `AuthForwardHandler` accordingly.
I've tested locally with this config https://gist.github.com/MVanderlee/2dba10f1ed6c869630eab27847bc2d12

The endpoint only has to support GET but wasn't sure how that's achieved. 

Please review carefully as I'm not familiar with GoLang.

### DIFF
--- a/api/decision.go
+++ b/api/decision.go
@@ -31,8 +31,8 @@ import (
 )
 
 const (
-	DecisionPath 				= "/decisions/generic"
-	LegacyDecisionPath 	= "/decisions"
+	DecisionPath       = "/decisions/generic"
+	LegacyDecisionPath = "/decisions"
 )
 
 type decisionHandlerRegistry interface {
--- a/api/decision.go
+++ b/api/decision.go
@@ -31,7 +31,8 @@ import (
 )
 
 const (
-	DecisionPath = "/decisions/generic"
+	DecisionPath 				= "/decisions/generic"
+	LegacyDecisionPath 	= "/decisions"
 )
 
 type decisionHandlerRegistry interface {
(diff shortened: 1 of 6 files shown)


MichielVanderlee:
I can't figure out how to get the CLA working. I have however accepted it

MichielVanderlee:
Just realized that the URI is only the path. Would also need to get add these:
```
	xForwardedProto  = "X-Forwarded-Proto"
	xForwardedHost   = "X-Forwarded-Host"
```

and somehow combine them into a URL object for the Matcher to understand

aeneasr:
Thank you for this PR. Obviously we'd like to avoid having one endpoint per proxy. Is there any other solution or workaround for this? If not, I think it would make sense to rewrite the decision endpoint and add e.g. `/decision/generic`, `/decision/traefik` - that ways it's really easy to target one individual service.

Regarding the CLA, I'll look into it but sometimes the thing is buggy.

MichielVanderlee:
There is no alternative from Traefik's side. I agree with the decision endpoint rewrite.

MichielVanderlee:
I've updated the paths, and now include the protocol and host in the matcher url.

api/decision.go
```diff
@@ -30,7 +30,7 @@ import (
 )
 
 const (
-	DecisionPath = "/decisions"
+	DecisionPath = "/decisions/generic"
```
aeneasr: We need to set up a redirect from `/decisions/*` (excluding "/decisions/generic(|/*)") to `/decisions/generic/$1` for backwards compatibility!

aeneasr:
Once the redirect is set up, I believe that all tests will pass again! :)

aeneasr:
Sorry, accidental close.

derekbar90:
Any update on the status of this ticket?

aeneasr:
@MichielVanderlee are you still up for this contribution?

pdaher:
@jmt-vanderlee do you still plan to go forward with this PR?

derekbar90:
@jmt-vanderlee Are you still thinking of accepting the license agreement? Would love to see this hit production

rdehouss:
Hi @mvanderlee, I see that your gist is not available anymore.
Do you have another one?

mrkwtz:
> Thank you for this PR. Obviously we'd like to avoid having one endpoint per proxy. Is there any other solution or workaround for this? If not, I think it would make sense to rewrite the decision endpoint and add e.g. `/decision/generic`, `/decision/traefik` - that ways it's really easy to target one individual service.
> 
> Regarding the CLA, I'll look into it - sometimes the thing is buggy.

We like the idea to make the decisions endpoint more configurable (e.g. define the header from which it should source host, uri, etc).

We ran into the same issues with the nginx ingress controller and auth_request. Because we don't want to expose the decision endpoint via ingress we target the oathkeeper-api directly in the auth-url. However this leads oathkeeper to get his hostname as the host header (therefore it doesn't match the hostname in the rules anymore) and there is no way around this (when you want to keep using the "self-configurable" nginx controller).

aeneasr:
I might be able to tackle this and provide a generic structure for adding proxy specific decision endpoints

api/decision.go
mvanderlee: I've handled this by simply relying on the middleware order. 
This does present an issue for the metrics, and I'm not sure how to handle that.

mvanderlee:
Proof of signing the CLA:
image

aeneasr:
Thank you! The CLA is sometimes buggy :)

I take it the code is ready for review? :)

aeneasr:
Superseded by #486

Issue #257, opened 2019-09-19
Support fully both Oauth & JWT authenticator in access rule 


piotrmsc:
**Is your feature request related to a problem? Please describe.**

Currently, you can define Oauth2 & JWT authenticator together, however, `Authorization: Bearer` is common header scheme for those two authenticators. 
If you define an access rule with 2 authenticators where Oauth2 is first and a request is made with JWT, such request will be rejected as token introspection rejects presented JWT - and vice versa.

**Describe the solution you'd like**
Would be really nice to allow developers create an access rule which defines where to look for token header in case of authenticator JWT & Oauth **or different alternative** allow to specify authorization header scheme.

Example with header selection :

```
[{
  "id": "some-id",
  "upstream": {
    "url": "http://my-backend-service"
  },
  "match": {
    "url": "http://my-app/some-route",
    "methods": [
      "GET"
    ]
  },
  "authenticators": [{
    "handler": "jwt",
    "config": {
      "required_scope": ["scope-a", "scope-b"],
      "target_audience": ["aud-1"],
      "trusted_issuers": ["iss-1"],
      "tokenHeader" : "x-jwt-authorization"
    }
  },
 {
    "handler": "oauth2_introspection",
    "config": {
      "required_scope": ["scope-a", "scope-b"]
    }
  }
],
  "authorizer": { "handler": "allow" },
  "mutator": { "handler": "noop" }
}]
```
Example with authorization header scheme selection

```
[{
  "id": "some-id",
  "upstream": {
    "url": "http://my-backend-service"
  },
  "match": {
    "url": "http://my-app/some-route",
    "methods": [
      "GET"
    ]
  },
  "authenticators": [{
    "handler": "jwt",
    "config": {
      "required_scope": ["scope-a", "scope-b"],
      "target_audience": ["aud-1"],
      "trusted_issuers": ["iss-1"],
      "authorizationScheme" : "jwt"
    }
  },
 {
    "handler": "oauth2_introspection",
    "config": {
      "required_scope": ["scope-a", "scope-b"]
    }
  }
],
  "authorizer": { "handler": "allow" },
  "mutator": { "handler": "noop" }
}]
```

**Additional context**

Additional considerations : 

- if access rules is defined with JWT/oauth  authenticator with custom header but such header is not present in the request (even not in `Authorization` header) oathkeeper should reject such call

[...]

PR #271, opened 2019-10-02
Support alternative token location

kubadz:
## Related issue

#257 

## Proposed changes

Add additional optional configuration to jwt and oauth2_introspection authenticators allowing to set from where (which header or query parameter) the token should be received. The configuration is a `token_from` field in per-rule-configuration, as described in a linked issue.

## Checklist

      the changes.
- [x] I have documented my changes in the

### DIFF
--- a/helper/bearer.go
+++ b/helper/bearer.go
@@ -25,14 +25,28 @@ import (
 	"strings"
 )
 
+const (
+	defaultAuthorizationHeader = "Authorization"
+)
+
 type BearerTokenLocation struct {
 	Header         *string `json:"header"`
 	QueryParameter *string `json:"query_parameter"`
 }
 
 func BearerTokenFromRequest(r *http.Request, tokenLocation *BearerTokenLocation) string {
-	auth := r.Header.Get("Authorization")
-	split := strings.SplitN(auth, " ", 2)
+	var token string
+	if tokenLocation != nil {
+		if tokenLocation.Header != nil {
+			token = r.Header.Get(*tokenLocation.Header)
+		} else if tokenLocation.QueryParameter != nil {
+			token = r.FormValue(*tokenLocation.QueryParameter)
+		}
+	} else {
+		token = r.Header.Get(defaultAuthorizationHeader)
+	}
+
+	split := strings.SplitN(token, " ", 2)
 	if len(split) != 2 || !strings.EqualFold(split[0], "bearer") {
 		return ""
 	}
--- a/helper/bearer.go
+++ b/helper/bearer.go
@@ -35,19 +35,14 @@ type BearerTokenLocation struct {
 }
 
 func BearerTokenFromRequest(r *http.Request, tokenLocation *BearerTokenLocation) string {
-	var token string
 	if tokenLocation != nil {
 		if tokenLocation.Header != nil {
-			token = r.Header.Get(*tokenLocation.Header)
+			return r.Header.Get(*tokenLocation.Header)
 		} else if tokenLocation.QueryParameter != nil {
-			token = r.FormValue(*tokenLocation.QueryParameter)
-		} else {
-			token = r.Header.Get(defaultAuthorizationHeader)
+			return r.FormValue(*tokenLocation.QueryParameter)
 		}
-	} else {
-		token = r.Header.Get(defaultAuthorizationHeader)
 	}
-
+	token := r.Header.Get(defaultAuthorizationHeader)
 	split := strings.SplitN(token, " ", 2)
 	if len(split) != 2 || !strings.EqualFold(split[0], "bearer") {
 		return ""
--- a/helper/bearer.go
+++ b/helper/bearer.go
@@ -25,7 +25,12 @@ import (
 	"strings"
 )
 
-func BearerTokenFromRequest(r *http.Request) string {
+type BearerTokenLocation struct {
+	Header         *string `json:"header"`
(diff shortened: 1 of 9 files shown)


helper/bearer.go
```diff
@@ -25,9 +25,28 @@ import (
...
+const (
+	defaultAuthorizationHeader = "Authorization"
+)
+
+type BearerTokenLocation struct {
+	Header         *string `json:"header"`
+	QueryParameter *string `json:"query_parameter"`
+}
+
+func BearerTokenFromRequest(r *http.Request, tokenLocation *BearerTokenLocation) string {
```
aeneasr: Would it make sense to test this independently?

kubadz: Yeah. That might be a good idea :)
I added tests.

helper/bearer.go
```diff
@@ -25,9 +25,28 @@ import (
...
+	var token string
+	if tokenLocation != nil {
+		if tokenLocation.Header != nil {
+			token = r.Header.Get(*tokenLocation.Header)
```
aeneasr: Wouldn't it make more sense to return the value here directly?

```suggestion
			return r.Header.Get(*tokenLocation.Header)
```

helper/bearer.go
```diff
@@ -25,9 +25,28 @@ import (
...
+		} else if tokenLocation.QueryParameter != nil {
+			token = r.FormValue(*tokenLocation.QueryParameter)
```
aeneasr: same here:

```suggestion
			return r.FormValue(*tokenLocation.QueryParameter)
```

helper/bearer.go
aeneasr: Wouldn't it make more sense to return the value here directly?

```suggestion
			return r.Header.Get(*tokenLocation.Header)
```

helper/bearer.go
```diff
@@ -25,9 +25,28 @@ import (
...
+		}
```
aeneasr: Is it possible that both if / else if do not get executed? Then `token` would be empty.

helper/bearer.go
kubadz: We can't return the value directly there, as in your suggestion, because we need to return only the token without the "bearer" part. However, I thought about moving lines 49-53 to a separate function. Then the line of your suggestion would look like:
```
return getToken(r.Header.Get(*tokenLocation.Header))
```

kubadz: What do you think?

helper/bearer.go
kubadz: That is not possible, as JSON schema for `jwt` and `oauth2_introspection` authenticators would not allow for that.

helper/bearer.go
aeneasr: The `bearer` is only relevant when the header is `Authorization` as as means of switching authorization modes (e.g. basic, bearer, ...).

However, an explicit query parameter such as `?token=1234` would not include the bearer instruction as the parameter is already explicitly bound to a token. The same applies if we define a *custom* header such as `X-MyApplication-AuthToken`. Here too, the context is explicit and clear.

helper/bearer.go
aeneasr: Can we make this explicit in the code by simply doing a if/else?

kubadz: We can, but if somehow QueryParameter would be a nil, than the program will crash on dereferencing a nil pointer. With "else if" the function will simply return an empty string, like if there was no token in the configured header/query parameter (what may be a proper behavior if there was "alternative token location" set, but not configured - even though with the current schema it is not possible).

I see an alternative solution here: returning an error saying that rule  is misconfigured. But I would have to add error handling in the places where this function is called. 

What do you think?

aeneasr: Ok, how about simply falling back to the authorization header in the case where both are nil?

helper/bearer.go
piotrmsc: I think in query param schema should not be present in the value but in custom header it coud be, this gives more guarantee/flexibility that some libs in developer  application will still work. OAuth for example expects `Bearer` scheme

kubadz: We can also make that optional for Header. So both `bearer token` and `token` will work. @aeneasr how would you prefere?

helper/bearer.go
kubadz: Ok, done

helper/bearer.go
aeneasr: > I think in query param schema should not be present in the value but in custom header it coud be, this gives more guarantee/flexibility that some libs in developer application will still work. OAuth for example expects Bearer scheme

That would be two authenticators with two separate custom headers IMO

piotrmsc: @aeneasr We do not give the possibility to override the scheme and IMO it should remain as it is but with the option to use different header

kubadz: @aeneasr the last word is yours. Do you want the "bearer" in custom header value to be optional, required or not valid?

aeneasr: I think it's very confusing to have the "bearer" in the custom header and the query parameter. If you look at the documentation it also does not become very clear. This is something I believe will make things very awkward when people are trying to use these features as I've never ever in my life have seen something like `X-Token-Auth: bearer bla` nor `?token=bearer%20bla`. So please remove the "bearer" splitting for these two parameters and, if unclear, explicitly document it with examples in the docs.

kubadz: Ok, done 😄

Issue #276, opened 2019-10-16
Env vars for jwks_url in id_token mutator not working in versions >v0.19.0-beta.1


kubadz:
**Describe the bug**

Mutator `id_token` requires `jwks_url` to be configured either in global or in per-rule configuration. The global configuration in most cases can be set by env variables. As the version v0.19.0 introduced changes in config (https://github.com/ory/oathkeeper/pull/258) it is logical that also env variables might have been changed.

Previously our configuration was setting `MUTATORS_ID_TOKEN_JWKS_URL` and it obviously didn't work in the recent oathkeeper version. I set it to `MUTATORS_ID_TOKEN_CONFIG_JWKS_URL` as it is currently configured [on master](https://github.com/ory/oathkeeper/blob/fc85ac854c3fb4cdd96bbae650f7355400431eac/driver/configuration/provider_viper.go#L62) but it also didn't work.

For both cases Oathkeeper returns error
```
{"error":{"code":500,"status":"Internal Server Error","request":"7da5f8a3-8db8-45dd-b358-e7eec13cdaaa","reason":"Configuration for mutator \"id_token\" could not be validated: jwks_url is required","message":"mutator matching this route is misconfigured or disabled"}}
```
when calling a service.

**some thoughts on this issue**: I suspect that it might be related to the way how config for mutators is validated. Maybe the validation with json schema happens on the config object without jwks_url set to value from env variable. I haven't checked it thoroughly in the code, though.

**Reproducing the bug**

1. Run Oathkeeper with env var `MUTATORS_ID_TOKEN_CONFIG_JWKS_URL` set to the jwks location, and with id_token mutator in global config set as following:
```
...
      id_token:
        config:
          issuer_url: https://example.com/
          ttl: 60s
...
```
2. Create a rule for some service with mutators configured as following:
```
        "mutators": [
          {
            "handler": "id_token",
            "config": {
              "issuer_url": "https://example.com/v2"
            }
          }
        ]
```
3. Create a request to that service. The response should contain an error mentioned in the description.

_Server logs_

```
time="2019-10-16T09:43:59Z" level=info msg="started handling request" method=GET remote="127.0.0.1:38458" request=/ request_id=98490e37-3176-4905-a407-83a24781921a
[...]

Issue #270, opened 2019-09-30
Env vars for client ID/secret in oauth2_introspection don't work anymore in v0.19.0-beta.1


GuillaumeSmaha:
**Describe the bug**

Setting environment vars `AUTHENTICATORS_OAUTH2_INTROSPECTION_PRE_AUTHORIZATION_CLIENT_ID` and `AUTHENTICATORS_OAUTH2_INTROSPECTION_PRE_AUTHORIZATION_CLIENT_SECRET` doesn't work anymore since `v0.19.0-beta.1`

If I don't put `client_id` and `client_secret` in global configuration, none error is throw at start for the configuration file but the rules will throw an error without information:
`time="2019-09-30T14:47:43Z" level=error msg="A rule uses a malformed configuration and all URLs matching this rule will not work. You should resolve this issue now." error="authenticator matching this route is misconfigured or disabled"` 

Global configuration for `oauth2_introspection`:
```yaml
    authenticators:
      oauth2_introspection:
        enabled: true
        config:
          introspection_url: https://
          scope_strategy: none
          required_scope: []
          pre_authorization:
            enabled: true
            token_url: https://
            scope: []
```

Example of rule:
```yaml
    - id: api
      upstream:
        url: http://backend
        preserve_host: true
      match:
        url: http://domain.com/api
        methods:
        - GET
        - POST
        - PUT
        - DELETE
        - PATCH
        - HEAD
        - OPTIONS
      authenticators:
      - handler: oauth2_introspection
      authorizer:
        handler: allow
```

**Expected behavior**

Env vars should be allowed for client id and secret to avoid saving secret in git repo.

**Environment**

- Version:`oryd/oathkeeper:v0.19.0-beta.1`

**Additional context**

I think this is related to this commit and lines removal:
https://github.com/ory/oathkeeper/commit/89709aabfe002fc5ae2e76016fc45a13d74f3d8b#diff-2449a6ea083767b149400840c05f41bdL58-L59

koenmtb1:
Can confirm. I'm having this issue too.

PR #279, opened 2019-10-21
cmd: Run BindEnv() on boot

aeneasr:
Closes #276 
Closes #270

Additionally, lines https://github.com/ory/oathkeeper/commit/89709aabfe002fc5ae2e76016fc45a13d74f3d8b#diff-2449a6ea083767b149400840c05f41bdL43-L74 should be added again.

Also, we need to document changes to the config keys (everything is now prefixed with `CONFIG`).

PR #280, opened 2019-10-21
Fix configuration with env vars

hypnoglow:
## Related issue

- https://github.com/ory/oathkeeper/issues/276
- https://github.com/ory/oathkeeper/issues/270

## Proposed changes

Currently env vars do not work for any configuration property. One reason is because [BindEnv](https://github.com/ory/oathkeeper/blob/40e3b891b99a41bee4b7be1a2cf7463bfb64f8db/driver/configuration/provider_viper.go#L90) is never called. The other thing that actually the usage of BindEnv is incorrect: this function DOES NOT accept multiple keys (see [source](https://github.com/spf13/viper/blob/master/viper.go#L960) for details). Even if BindEnv would be used correctly, it will not resolve env vars like `SERVE_PROXY_PORT` because it does not replaces dots with underscores.

This MR provides the same way of loading env vars as in [hydra](https://github.com/ory/hydra/blob/master/driver/configuration/provider_viper.go#L72).

## Checklist

      the changes.

### DIFF
--- a/driver/configuration/provider_viper.go
+++ b/driver/configuration/provider_viper.go
@@ -87,7 +87,7 @@ const (
 )
 
 func BindEnvs() {
-	if err := viper.BindEnv(
+	keys := []string{
 		ViperKeyProxyReadTimeout,
 		ViperKeyProxyWriteTimeout,
 		ViperKeyProxyIdleTimeout,
@@ -112,8 +112,13 @@ func BindEnvs() {
 		ViperKeyAuthenticatorOAuth2ClientCredentialsIsEnabled,
 		ViperKeyAuthenticatorOAuth2TokenIntrospectionIsEnabled,
 		ViperKeyAuthenticatorUnauthorizedIsEnabled,
-	); err != nil {
-		panic(err.Error())
+	}
+
+	for _, key := range keys {
+		env := strings.ToUpper(strings.ReplaceAll(key, ".", "_"))
+		if err := viper.BindEnv(key, env); err != nil {
+			panic(err.Error())
+		}
 	}
 }
 
--- a/driver/configuration/provider_viper_public_test.go
+++ b/driver/configuration/provider_viper_public_test.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/url"
+	"os"
 	"testing"
 
 	"github.com/rs/cors"
@@ -341,4 +342,22 @@ func TestAuthenticatorOAuth2TokenIntrospectionPreAuthorization(t *testing.T) {
 			}
 		})
 	}
+}
+
+func TestViperProvider_envs(t *testing.T) {
+	p := NewViperProvider(logrus.New())
+	BindEnvs()
+
+	assert.Equal(t, ":4455", p.ProxyServeAddress())
+	assert.Equal(t, false, p.AuthenticatorIsEnabled("oauth2_introspection"))
+
+	err := os.Setenv("SERVE_PROXY_HOST", "foo")
+	require.NoError(t, err)
+	err = os.Setenv("SERVE_PROXY_PORT", "8080")
+	require.NoError(t, err)
+	err = os.Setenv("AUTHENTICATORS_OAUTH2_INTROSPECTION_ENABLED", "true")
+	require.NoError(t, err)
+
+	assert.Equal(t, "foo:8080", p.ProxyServeAddress())
+	assert.Equal(t, true, p.AuthenticatorIsEnabled("oauth2_introspection"))
 }
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -30,11 +30,10 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/ory/gojsonschema"
-
+	"github.com/ory/oathkeeper/driver/configuration"
 	"github.com/ory/viper"
-	"github.com/ory/x/viperx"
-
 	"github.com/ory/x/logrusx"
+	"github.com/ory/x/viperx"
 )
 
 var schemas = packr.New("schemas", "../.schemas")
@@ -63,6 +62,7 @@ func init() {
 	}
 
 	cobra.OnInitialize(func() {
+		configuration.BindEnvs()
 		viperx.InitializeConfig("oathkeeper", "", nil)
 
 		logger = logrusx.New()


driver/configuration/provider_viper.go
```diff
@@ -86,35 +86,9 @@ const (
...
-		ViperKeyAuthenticatorJWTIsEnabled,
-		ViperKeyAuthenticatorOAuth2ClientCredentialsIsEnabled,
-		ViperKeyAuthenticatorOAuth2TokenIntrospectionIsEnabled,
-		ViperKeyAuthenticatorUnauthorizedIsEnabled,
-	); err != nil {
-		panic(err.Error())
-	}
+func init() {
+	viper.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
+	viper.AutomaticEnv()
```
aeneasr: This is executed properly by `viperx` on boot. No need to set/use this here.

driver/configuration/provider_viper.go
```diff
@@ -86,35 +86,9 @@ const (
 	ViperKeyAuthenticatorUnauthorizedIsEnabled = "authenticators.unauthorized.enabled"
 )
 
-func BindEnvs() {
-	if err := viper.BindEnv(
-		ViperKeyProxyReadTimeout,
```
aeneasr: Removing these will have a detrimental effect. Viper does not automatically check environment variables. You have to bind them explicitly.

driver/configuration/provider_viper_public_test.go
```diff
@@ -342,3 +341,18 @@ func TestAuthenticatorOAuth2TokenIntrospectionPreAuthorization(t *testing.T) {
...
+	assert.Equal(t, ":4455", p.ProxyServeAddress())
+	assert.Equal(t, false, p.CORSEnabled("proxy"))
+
+	err := os.Setenv("SERVE_PROXY_PORT", "8080")
+	require.NoError(t, err)
+	err = os.Setenv("SERVE_PROXY_CORS_ENABLED", "true")
+	require.NoError(t, err)
+
+	assert.Equal(t, ":8080", p.ProxyServeAddress())
+	assert.Equal(t, true, p.CORSEnabled("proxy"))
```
aeneasr: This test makes a lot of sense! However, it should be changed to use `BindEnv()` and we should test config keys such as OAuth2 Client ID. Some keys, such as CORS, are set using libraries iirc.

driver/configuration/provider_viper.go
hypnoglow: I'm not sure why it is called in `InitializeConfig` and not either in the init function or the separate function. This way tests that simply call `NewViperProvider` are never getting this effect. I think this should be encapsulated in the provider. Maybe even still in both places.

driver/configuration/provider_viper.go
hypnoglow: > Viper does not automatically check environment variables. You have to bind them explicitly.

Why not? It indeed loads envs automatically: https://github.com/spf13/viper/blob/master/viper.go#L1039

In hydra, everything works fine without any explicit bindings.

driver/configuration/provider_viper.go
aeneasr: Because we're initializing viper the same way in all projects. Encapsulating it in a library helps with keeping things coherent between projects. This will not change.

These configs are not useful for testing, we do not set env vars in such a way in our tests. For isolated tests, like the one you did, we can manually either invoke those two functions or export the functionality in `viperx`.

driver/configuration/provider_viper.go
aeneasr: Your analysis is unfortunately incorrect. In Hydra, we do not (yet) use JSON Schema to validate the config. However, validating against the Config JSON Schema requires all keys to be loaded. That's just the way Viper works (it is quirky).

hypnoglow: Sorry, I do not understand how we suppose to explicitly bind all keys, since most of them were recently removed for example in https://github.com/ory/oathkeeper/pull/258 . How do envs like `AUTHENTICATORS_OAUTH2_INTROSPECTION_CONFIG_PRE_AUTHORIZATION_CLIENT_ID` suppose to work?

(also related to the comment below about tests)

aeneasr: While the keys have been removed as constants (because we don't explicitly need them any more - except for `BindEnvs()`). However, I think it would be better to add a patch to our fork to automatically bind all envs when doing `AutomaticEnv()` instead of doing both `AutomaticEnv()` as well as `BindEnv()`. This is a known (and unsolved issue) in viper that has lead to some frustration (me included): https://github.com/spf13/viper/issues/761 (and all linked issues).

Then, we can throw away `BindEnvs()` which - to be honest - sucks anyways.

aeneasr: This however would need some thinking as `_` can both be `.` as well as `_`. So we'd end up with a lot of keys when "autom-mounting" all environment keys. For example env var `FOO_BAR` could both be `foo_bar` as well as `foo.bar`. That would in turn cause issues when we set `additionalProperties: false` in the JSON Config Schema - which is not great because with that property set people would immediately find any typos they're doing. Maybe by defining some "categories" that always replace `_` with `.` would be the way to go. Not sure yet, needs some thinking.

aeneasr: Actually, I just found my original issue which touched this topic in 2016 :D https://github.com/spf13/viper/issues/212

hypnoglow: Oh, I see, things getting complicated, combining viper behaviour and the introduction of json schemas.

If you come up with the idea how should we solve it, let me know.

I've pushed changes that make env vars work for defined keys (e.g. `SERVE_PROXY_PORT`), but envs like `AUTHENTICATORS_OAUTH2_INTROSPECTION_CONFIG_PRE_AUTHORIZATION_CLIENT_ID` still do not work. If you feel like we don't need this intermediary improvement, feel free to close this MR.

aeneasr:
I wrote a small helper that is capable of extracting all configuration keys from the JSON Schema and that binds those keys to viper. This will resolve all of #280 #279 #276 #270 in a way that's future proof and very little effort code-wise and further homogenizes the configuration setup: https://github.com/ory/x/pull/85

hypnoglow:
@aeneasr Looks like a nice solution to the problem! Thanks

PR #291, opened 2019-11-01
Fix always recompile regex URL for rules

lehainam-dev:
## Related issue

This [condition](https://github.com/ory/oathkeeper/blob/master/rule/rule.go#L178) always return `true`, because `Rule.Match.compiledURL` and `Rule.Match.compiledURLChecksum` is never updated in `RepositoryMemory.rules`, leading the expensive function `CompileRegex` always call each time `oathkeeper` check an URL is match rules or not.

I found this issue when testing `oathkeeper` performance with `pprof`.

## Proposed changes

I change the way `RepositoryMemory.rules` and `Rule.Match` should update. 

Honestly, I am not familiar with golang much. Please review and suggest if you have are better solutions.

## Checklist

      the changes.

### DIFF
--- a/rule/repository_memory.go
+++ b/rule/repository_memory.go
@@ -109,10 +109,11 @@ func (m *RepositoryMemory) Match(ctx context.Context, method string, u *url.URL)
 	defer m.RUnlock()
 
 	var rules []Rule
-	for _, rule := range m.rules {
+	for i, rule := range m.rules {
 		if err := rule.IsMatching(method, u); err == nil {
 			rules = append(rules, rule)
 		}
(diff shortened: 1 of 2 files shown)


aeneasr:
Thank you! Would it be possible to add a test so this doesn't break in a future change accidentally?

aeneasr:
Just checking back in if you'd be open for adding a test

lehainam-dev:
Hi @aeneasr ,
Sorry, I don't think that I would be available to write some tests.
I will pass this issue to your team. Thanks for your incredible work.

Issue #298, opened 2019-11-14
cookie_session authenticator masks 401 errors


paulbdavis:
**Describe the bug**

When the endpoint called by the `cookie_session` checker returns a 401, the proxy returns a 403

**Reproducing the bug**

1. Set up a `cookie_session` authenticator with a service that will return a 401 when the cookie in question is invalid
2. Attempt to access an endpoint using that authenticator

**Expected behavior**

I would expect the authenticator to return a 401 when the session checker returns a 401

**Environment**

- Version: v0.32.1-beta.1
- Environment: kubernetes

aeneasr:
This is resolved on master by #315

PR #300, opened 2019-11-15
handle 401 when checking cookie session

paulbdavis:
## Related issue

#298 

## Proposed changes

Handles 401 errors when attempting to check a cookie session rather than masking it as a 403 error

## Checklist

      the changes.

### DIFF
--- a/pipeline/authn/authenticator_cookie_session.go
+++ b/pipeline/authn/authenticator_cookie_session.go
@@ -66,7 +66,7 @@ func (a *AuthenticatorCookieSession) Authenticate(r *http.Request, config json.R
 	origin := cf.CheckSessionURL
 	body, err := forwardRequestToSessionStore(r, origin)
 	if err != nil {
-		return nil, helper.ErrForbidden.WithReason(err.Error()).WithTrace(err)
+		return nil, err
 	}
 
 	var session struct {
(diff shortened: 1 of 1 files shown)


pipeline/authn/authenticator_cookie_session.go
```diff
@@ -115,10 +115,12 @@ func forwardRequestToSessionStore(r *http.Request, checkSessionURL string) (json
 	if res.StatusCode == 200 {
 		body, err := ioutil.ReadAll(res.Body)
 		if err != nil {
-			return json.RawMessage{}, err
+			return json.RawMessage{}, helper.ErrForbidden.WithReason(err.Error()).WithTrace(err)
```
aeneasr: - Why would you consider the inability to read the response body a 401 error? Wouldn't it make more sense to return more error details here? I believe this will be incredibly hard to handle. I understand that this was the case before, but since we're working on this patch we can address it straight away.
- `.WithTrace()` adds the stack trace from the upstream error to the error context. stdlib errors do not implement stack trace functionality, therefore `WithTrace()` on a stdlib doesn't have an effect.
- Without `errors.WithStack()` it's not possible to get the error trace. What was your thinking behind omitting that?

pipeline/authn/authenticator_cookie_session.go
```diff
@@ -115,10 +115,12 @@ func forwardRequestToSessionStore(r *http.Request, checkSessionURL string) (json
...
 		}
 		return json.RawMessage(body), nil
-	} else {
-		return json.RawMessage{}, errors.WithStack(helper.ErrUnauthorized)
+	} else if res.StatusCode == 401 {
+		return json.RawMessage{}, helper.ErrUnauthorized
 	}
+
+	return json.RawMessage{}, errors.WithStack(helper.ErrUnauthorized)
```
aeneasr: These changes do not do a lot except changing some `errors.WithStack()` behavior, which should not have any effect.

pipeline/authn/authenticator_cookie_session.go
paulbdavis: I added that here because the function that called `forwardRequestToSessionStore` was previously wrapping all the errors in a 403, so I added this to maintain the existing behavior for all of the other error cases, though I would say this should probably be a 500 error instead

aeneasr: Ok, so let's return instead:

```suggestion
			return json.RawMessage{}, errors.WithStack(helper.ErrForbidden.WithReasonf("Unable to fetch cookie session context from remote: %+v", err))
```

aeneasr:
Are you still up for these changes? :)

paulbdavis:
I think the change makes sense in general, but I'm not using the cookie auth anymore since the patch to use cookies for oauth was merged, so it's not a priority for me anymore.

PR #301, opened 2019-11-16
add cookie as an option for oauth2_introspection authenticator

paulbdavis:
## Proposed changes

Adds the ability to get the OAuth2 bearer token from a cookie in addition to header or query param. This allows an http-only, secure cookie to be used for storing oauth credentials.

## Checklist

      the changes.
- [x] I have documented my changes in the

## Further comments

I'm not quite sure how to properly handle the schema files, I'm guessing there is some compilation going on there that is not documented (vs just editing the `config.schema.json` file as I have for now).

### DIFF
--- a/.schemas/config.schema.json
+++ b/.schemas/config.schema.json
@@ -502,7 +502,7 @@
                 "cookie"
               ],
               "properties": {
-                "query_parameter": {
+                "cookie": {
                   "title": "Cookie",
                   "type": "string",
                   "description": "The cookie (case sensitive) that must contain a token for request authentication.\n It can't be set along with header or query_parameter."
--- a/.schemas/config.schema.json
+++ b/.schemas/config.schema.json
@@ -481,7 +481,7 @@
                 "header": {
                   "title": "Header",
                   "type": "string",
-                  "description": "The header (case insensitive) that must contain a token for request authentication.\n It can't be set along with query_parameter."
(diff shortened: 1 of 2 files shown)


.schemas/config.schema.json
```diff
@@ -493,7 +493,19 @@
...
+                  "description": "The query parameter (case sensitive) that must contain a token for request authentication.\n It can't be set along with header or cookie."
+                }
+              }
+            },
+            {
+              "required": [
+                "cookie"
+              ],
+              "properties": {
+                "query_parameter": {
```
aeneasr: I think this should be `cookie` right?

paulbdavis:
> I'm saying that because many people think that cookies are "secure" and we should make sure to document this appropriately to avoid any misconceptions.

@aeneasr I meant the "secure" flag is set on the cookie, not that cookies are inherently secure.

As with the cookie_session handler, the onus to protect against CSRF and XSS is on the application developer, not on oathkeeper, correct?

.schemas/config.schema.json
paulbdavis: Yes it should, fixed and pushed

paulbdavis:
For this and #302, I can also go ahead and add the same options to other oauth2 authenticator before it gets merged, just so those two are at feature parity

aeneasr:
Makes sense!

paulbdavis:
Actually, after review, it looks like these options are not really appropriate for the client_credentials authenticator. I think this and #302 are ready to be merged as is

PR #318, opened 2019-12-17
Switch to alpine Docker base-image

### LINKED ISSUE #312: Replace the base image with alpine
In k8s, `scratch` mirroring is very inconvenient. You cannot execute commands such as mount ls, and you cannot debug the problem.

aeneasr:
Closes #312

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -72,12 +72,24 @@ scoop:
     email: hi@ory.sh
 
 dockers:
-  - image_templates:
+  - dockerfile: Dockerfile
+    binaries:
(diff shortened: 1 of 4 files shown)


tleef:
Does this have a significant impact on the image size or cpu/memory usage at runtime?

aeneasr:
no

aeneasr:
Sorry, I didn't see the edit:

> I noticed that you have a separate Dockerfile-alpine for hydra. Maybe that could be the way to go here as well?

Yes, that's absolutely right and I've updated the PR accordingly! :)

Issue #324, opened 2019-12-27
Clean up docker compose file


aeneasr:
The current docker-compose file is very outdated and should be updated or removed.

KarthikNayak:
Picking this up! Will update the compose file!

KarthikNayak:
@aeneasr I see that the `Docker` file doesn't contain the build instructions, instead the build instructions lay in the `Makefile` this also means that just running `docker build .` fails. Is this expected behavior? 
If so, is it okay to move the build instruction to the `Dockerfile` ? This would get the `docker-compose` up and running.

aeneasr:
Thank you @KarthikNayak ! Yeah right - please leave the current Docker Images as is! You can however add another Dockerfile named `Dockerfile-dc` with build instructions specifically for Docker Compose!

PR #325, opened 2019-12-30
Docker compose cleanup

KarthikNayak:
## Related issue

https://github.com/ory/oathkeeper/issues/324

## Proposed changes

Cleanup the existing docker-compose.yml and add a new Dockerfile-dc

## Checklist

      the changes.

### DIFF
--- a/Dockerfile-dc
+++ b/Dockerfile-dc
@@ -1,0 +1,18 @@
+FROM golang:1.13-alpine
+
+RUN addgroup -S ory; \
+    adduser -S ory -G ory -D -H -s /bin/nologin
+
+RUN apk add -U --no-cache ca-certificates
+
+ADD . /app
+WORKDIR /app
+ENV GO111MODULE on
+RUN go get -u github.com/gobuffalo/packr/v2/packr2
+RUN packr2
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build
+
+USER ory
+
+ENTRYPOINT ["/app/oathkeeper"]
+CMD ["serve"]
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,62 +1,17 @@
-version: '2'
+version: '3.7'
 
 services:
-
-  oathkeeper-migrate:
+  oathkeeper:
     build:
       context: .
-      dockerfile: Dockerfile
-    links:
-      - postgresd:postgresd
-    environment:
-      - LOG_LEVEL=debug
-    command:
-      migrate sql postgres://dbuser:secret@postgresd:5432/accesscontroldb?sslmode=disable
-    restart: on-failure
-
-  oathkeeper-proxy:
-    build:
-      context: .
-      dockerfile: Dockerfile
-    links:
-      - postgresd:postgresd
+      dockerfile: Dockerfile-dc
     ports:
       - "4455:4455"
-    depends_on:
-      - oathkeeper-api
-    command:
-      serve proxy
-    environment:
-      - LOG_LEVEL=debug
-      - PORT=4455
-      - ISSUER_URL=http://localhost:4455/
-      - OATHKEEPER_API_URL=http://oathkeeper-api:4456
-      - CREDENTIALS_ISSUER_ID_TOKEN_HS256_SECRET=arandomsecretarandomsecretarando
-    restart: on-failure
-
-  oathkeeper-api:
-    build:
-      context: .
-      dockerfile: Dockerfile
-    links:
-      - postgresd:postgresd
-    ports:
       - "4456:4456"
-    depends_on:
-      - oathkeeper-migrate
     command:
-      serve api
-    environment:
-      - LOG_LEVEL=debug
-      - PORT=4456
-      - DATABASE_URL=postgres://dbuser:secret@postgresd:5432/accesscontroldb?sslmode=disable
-      - ISSUER_URL=http://localhost:4455/
-      - CREDENTIALS_ISSUER_ID_TOKEN_HS256_SECRET=arandomsecretarandomsecretarando
+      serve --config=/etc/config/oathkeeper/config.yaml
+    volumes:
+      - type: bind
+        source: ./.docker_compose
+        target: /etc/config/oathkeeper
     restart: on-failure
-
-  postgresd:
-    image: postgres:9.6
-    environment:
-      - POSTGRES_USER=dbuser
-      - POSTGRES_PASSWORD=secret
-      - POSTGRES_DB=accesscontroldb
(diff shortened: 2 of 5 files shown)


ducharmemp:
I believe that this might also need a volume mount at $HOME/.oathkeeper.json or something like that? I'm actually in the process of trying to figure out how to set up my API access rules now and struggling with the fact that the full-stack example is also out of date.

I suppose another solution would be to build another image based off of oathkeeper with my config options, but that feels a bit heavy handed.

If the docker-compose isn't meant as an example just ignore me :)

KarthikNayak:
@ducharmemp I'll leave that decision to @aeneasr, currently this docker-compose uses the built in `.schemas` folder.

aeneasr:
Thank you, there is no need for any type of mounting. We're using packr as part of the build process to embed these files, as you can see here: https://github.com/ory/oathkeeper/blob/master/Makefile#L50-L56

Please update the docker image accordingly :)

KarthikNayak:
@aeneasr I've updated the PR, do have a look! :D

Dockerfile-dc
```diff
@@ -0,0 +1,21 @@
+FROM alpine:3.10
+
+RUN apk add -U --no-cache ca-certificates
+
+FROM golang:latest
+
+ADD . /app
+WORKDIR /app
+RUN go get -u github.com/gobuffalo/packr/v2/packr2
+RUN CGO_ENABLED=0 GO111MODULE=on GOOS=linux GOARCH=amd64 packr2
```
aeneasr: `packr2` does not need these arguments - see the makefile for reference.

Dockerfile-dc
```diff
@@ -0,0 +1,21 @@
+FROM alpine:3.10
+
+RUN apk add -U --no-cache ca-certificates
+
+FROM golang:latest
+
+ADD . /app
+WORKDIR /app
+RUN go get -u github.com/gobuffalo/packr/v2/packr2
```
aeneasr: We are using a fixed version of packr2, `GO111MODULE=on go install github.com/gobuffalo/packr/v2/packr2` should be enough - see the makefile for reference

Dockerfile-dc
```diff
@@ -0,0 +1,21 @@
+FROM alpine:3.10
```
aeneasr: Use `golang:1.13-alpine` instead. You don't need `golang:latest` then

docker-compose.yml
```diff
@@ -1,25 +1,10 @@
-version: '2'
```
aeneasr: This file needs more work - there is for example no config file being loaded. The environment variables also seem to be outdated

Dockerfile-dc
KarthikNayak: Will do!

Dockerfile-dc
KarthikNayak: Will do!

Dockerfile-dc
KarthikNayak: Alright, will change that!

docker-compose.yml
KarthikNayak: I somehow missed adding these changes to the commit, will add and push! :D

KarthikNayak: Alright, will check and add it!

KarthikNayak:
Just an FYI, got busy with life, picking this up again!

Dockerfile-dc
KarthikNayak: We would still need the GO111MODULE=on argument, this ensures that packr2 doesn't build the boxes as per the path instead of the modules.

aeneasr: right, that makes sense. You can set `ENV GO111MODULE on` in the dockerfile so go modules are enabled for all commands

aeneasr:
All good, happens to the best of us :)

Dockerfile-dc
KarthikNayak: Yup, we could do that too!

KarthikNayak:
@aeneasr You mentioned some of the ENV variables to be deprecated, I tried following https://www.ory.sh/docs/oathkeeper/configure-deploy, but the documentation seems to be outdated here too (it talks about using a DB, which we no longer require). Is there a more recent updated version of the documentation?

aeneasr:
Regarding docs, it was a bug that was introduced two days ago. It was resolved today so everything should be up to date again! :)

Dockerfile-dc
```diff
@@ -0,0 +1,23 @@
...
+ENV GO111MODULE on
+RUN go get -u github.com/gobuffalo/packr/v2/packr2
+RUN packr2
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build
+
+FROM scratch
+
+COPY --from=0 /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
+COPY --from=0 /app/oathkeeper /usr/bin/oathkeeper
+COPY --from=0 /app/.docker_compose/config.yaml config.yaml
```
aeneasr: As I said in the summary comment, adding config files as part of the build process is bad practice. Let's remote these. I've created another image for another project which makes use of volumes to mount the config. It uses a standard alpine-based image with a custom user: https://github.com/ory/kratos/blob/master/.docker/Dockerfile

Then, we simply [mount the folder where the config files are at](https://github.com/ory/kratos/blob/master/quickstart.yml#L82-L85) and use the [config flag](https://github.com/ory/kratos/blob/master/quickstart.yml#L75) to define where the config files is saved.

I think a 1:1 copy should work more or less for oathkeeper too. The cool thing is, hot reloading will work! So you can edit your config file on the host and oathkeeper will reload it!

Dockerfile-dc
```diff
@@ -0,0 +1,23 @@
...
+RUN apk add -U --no-cache ca-certificates
+
+ADD . /app
+WORKDIR /app
+ENV GO111MODULE on
+RUN go get -u github.com/gobuffalo/packr/v2/packr2
+RUN packr2
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build
+
+FROM scratch
```
aeneasr: I think we can use the alpine-based image here, as people might want to `docker exec` into it to debug or whatever. That way, they'll have a shell available!

Check out another docker image that uses an alpine-based linux and adds a user `ory`. I'd recommend doing something similar: https://github.com/ory/kratos/blob/master/.docker/Dockerfile

Dockerfile-dc
KarthikNayak: You're right, I'm just wondering if there's a better way to even add the private keys. I've made the changes you've suggested here.

Dockerfile-dc
KarthikNayak: Got rid of the multistage multi image Docker and using `golang:1.13-alpine`

Issue #330, opened 2020-01-03
support the cookie as token source (token_from) for JWT authenticator


ngrigoriev:
**Is your feature request related to a problem? Please describe.**

According to the Oathkeeper documentation the "jwt" authenticator can fetch the token from either a named HTTP header or from the query parameter. At the same time, "oauth2_introspection" authenticator supports the named cookie as source. It would be great to support the cookie source for "jwt" authenticator as well.

**Describe the solution you'd like**

I may be wrong but it looks like it may be already supported, authenticator_jwt.go calls helper.BearerTokenFromRequest() to get the token and this function does support all 3 sources. Is it possible that this is just a missed point in the documentation?

Thanks!

aeneasr:
Yes, I believe this is just a docs issue!

ngrigoriev:
I am trying to add the unit tests for this + update the schema

PR #331, opened 2020-01-07
pipeline/authn: add unit test for token_from->cookie for both jwt and oauth2_introspection authenticators (#330)

ngrigoriev:
pipeline/authn: add unit test for token_from->cookie for both jwt and oauth2_introspection authenticators (#330)

Update the schemas to add missing cookie config element

Signed-off-by: Grigoriev, Nikolai <ngrigoriev@gmail.com>

## Proposed changes

Add both negative and positive unit tests for both authenticators for "cookie" token origin

## Checklist

- [X] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md)
- [X] I have read the [security policy](../security/policy)
      the changes.
- [X] I have added tests that prove my fix is effective or that my feature works
- [x] Separate PR is provided for docs repo

### DIFF
--- a/.schemas/authenticators.jwt.schema.json
+++ b/.schemas/authenticators.jwt.schema.json
@@ -52,7 +52,7 @@
     },
     "token_from": {
       "title": "Token From",
-      "description": "The location of the token.\n If not configured, the token will be received from a default location - 'Authorization' header.\n One and only one location (header or query) must be specified.",
+      "description": "The location of the token.\n If not configured, the token will be received from a default location - 'Authorization' header.\n One and only one location (header, query, or cookie) must be specified.",
       "oneOf": [
         {
           "type": "object",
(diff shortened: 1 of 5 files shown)


aeneasr:
Nice! This looks good - as I mentioned in the docs PR I'd like the schema changes to be updated here as well, otherwise this is good to go!

ngrigoriev:
OK, I have verified the difference - for the affected section there was an empty line that I slipped into the schema in the docs repository. Since I have reverted the change there, this should be good, I think. BTW, I have noticed that the schema files in docs and here are slightly different, looks like the arrays are formatted differently.

aeneasr:
Ok perfect!

> BTW, I have noticed that the schema files in docs and here are slightly different, looks like the arrays are formatted differently.

Yeah we're prettifying the files in ory/docs which can lead to slightly different formatting :)

PR #334, opened 2020-01-08
#321

### LINKED ISSUE #321: Implement additional matching strategies and use negative lookahead regex
**Is your feature request related to a problem? Please describe.**

At the moment, access rules will match based on a regular expression, as documented [here (see bullet point `match`)](https://www.ory.sh/docs/oathkeeper/api-access-rules#access-rule-format).

There are two problems related to that:

1. The regex stdlib does not support negative lookahead, which can be cumbersome in real-world application as it's not possible to do a "match all except this one element".
2. Many developers do not like writing regular expressions, but are used to glob pattern matching instead.

**Describe the solution you'd like**

For **regex**, we should pick a different library (research needed as to which library suits the best) - one that supports negative lookahead/behind and replace the stdlib regex with that one. There should not be any backwards incompatible changes here!

For **glob matching**, we should use [gobwas/glob](https://github.com/gobwas/glob). Before implementing glob matching, we have to decide:

1. How will the access rule decide if glob or regex should be used?
2. What should the default behavior be glob or regex?

Once these decisions are made, we need to:

1. If the default behavior is glob, we need to implement an access [rule migrator](https://github.com/ory/oathkeeper/blob/master/rule/rule_migrator.go) that explicitly enables the "regex" matching strategy.
2. An abstraction needs to be implemented in the matching algorithm so that we can easily add new pattern matchers easily and without a lot of work.
3. These changes need to be documented.

ayzu:
## Related issue

Closes #321 

## Checklist

- [X ] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md)
- [X ] I have read the [security policy](../security/policy)
- [X ] I confirm that this pull request does not address a security
      the changes.
- [X ] I have added tests that prove my fix is effective or that my feature works
- [X ] I have added necessary documentation within the code base (if appropriate)
- [X ] I have documented my changes in the

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,8 @@
     <a href="https://discord.gg/PAMQWkr">Chat</a> |
     <a href="https://community.ory.sh/">Forums</a> |
     <a href="http://eepurl.com/di390P">Newsletter</a><br/><br/>
+    <a href="https://www.ory.sh/docs/guides/master/oathkeeper/">Guide</a> |
+    <a href="https://www.ory.sh/docs/oathkeeper/sdk/api">API Docs</a> |
     <a href="https://www.ory.sh/docs/oathkeeper/">Guide</a> |
     <a href="https://www.ory.sh/docs/oathkeeper/sdk/api">API Docs</a> |
     <a href="https://godoc.org/github.com/ory/oathkeeper">Code Docs</a><br/><br/>
--- a/README.md
+++ b/README.md
@@ -257,7 +257,6 @@ The Guide is available
 [here](https://www.ory.sh/docs/oathkeeper/).
 
 ### HTTP API documentation
-
 The HTTP API is documented
 [here](https://www.ory.sh/docs/oathkeeper/sdk/api).
 
--- a/README.md
+++ b/README.md
@@ -4,10 +4,8 @@
     <a href="https://discord.gg/PAMQWkr">Chat</a> |
     <a href="https://community.ory.sh/">Forums</a> |
     <a href="http://eepurl.com/di390P">Newsletter</a><br/><br/>
-    <a href="https://www.ory.sh/docs/guides/master/oathkeeper/">Guide</a> |
     <a href="https://www.ory.sh/docs/oathkeeper/sdk/api">API Docs</a> |
     <a href="https://www.ory.sh/docs/oathkeeper/">Guide</a> |
-    <a href="https://www.ory.sh/docs/oathkeeper/sdk/api">API Docs</a> |
     <a href="https://godoc.org/github.com/ory/oathkeeper">Code Docs</a><br/><br/>
     <a href="https://opencollective.com/ory">Support this project!</a>
 </h4>
@@ -257,6 +255,7 @@ The Guide is available
 [here](https://www.ory.sh/docs/oathkeeper/).
 
 ### HTTP API documentation
+
 The HTTP API is documented
 [here](https://www.ory.sh/docs/oathkeeper/sdk/api).
 
--- a/driver/configuration/provider.go
+++ b/driver/configuration/provider.go
@@ -19,6 +19,15 @@ const (
 	ForbiddenStrategyErrorType = "forbidden"
 )
 
+// MatchingStrategy defines matching strategy such as Regexp or Glob.
+type MatchingStrategy int
+
+// Possible values.
+const (
+	Regexp MatchingStrategy = iota
+	Glob
+)
+
 type Provider interface {
 	CORSEnabled(iface string) bool
 	CORSOptions(iface string) cors.Options
@@ -33,6 +42,7 @@ type Provider interface {
 	ProxyIdleTimeout() time.Duration
 
 	AccessRuleRepositories() []url.URL
+	AccessRuleMatchingStrategy() MatchingStrategy
 
 	ProxyServeAddress() string
 	APIServeAddress() string
--- a/driver/configuration/provider.go
+++ b/driver/configuration/provider.go
@@ -20,12 +20,13 @@ const (
 )
 
 // MatchingStrategy defines matching strategy such as Regexp or Glob.
-type MatchingStrategy int
+// Empty string defaults to "regexp"
+type MatchingStrategy string
 
-// Possible values.
+// Possible matching strategies.
 const (
-	Regexp MatchingStrategy = iota
-	Glob
+	Regexp MatchingStrategy = "regexp"
+	Glob   MatchingStrategy = "glob"
 )
 
 type Provider interface {
--- a/driver/configuration/provider.go
+++ b/driver/configuration/provider.go
@@ -20,7 +20,7 @@ const (
 )
 
 // MatchingStrategy defines matching strategy such as Regexp or Glob.
-// Empty string defaults to "regexp"
+// Empty string defaults to "regexp".
 type MatchingStrategy string
 
 // Possible matching strategies.
--- a/pipeline/authz/keto_engine_acp_ory.go
+++ b/pipeline/authz/keto_engine_acp_ory.go
@@ -94,6 +94,11 @@ func (a *AuthorizerKetoEngineACPORY) Authorize(r *http.Request, session *authn.A
 		return err
 	}
 
+	// only Regexp matching strategy is supported for now.
+	if !(a.c.AccessRuleMatchingStrategy() == "" || a.c.AccessRuleMatchingStrategy() == configuration.Regexp) {
+		return helper.ErrNonRegexpMatchingStrategy
+	}
+
 	subject := session.Subject
 	if cf.Subject != "" {
 		templateId := fmt.Sprintf("%s:%s", rule.GetID(), "subject")
--- a/pipeline/authz/keto_engine_acp_ory.go
+++ b/pipeline/authz/keto_engine_acp_ory.go
@@ -94,11 +94,6 @@ func (a *AuthorizerKetoEngineACPORY) Authorize(r *http.Request, session *authn.A
 		return err
 	}
 
-	compiled, err := rule.CompileURL()
-	if err != nil {
-		return errors.WithStack(err)
-	}
-
 	subject := session.Subject
 	if cf.Subject != "" {
 		templateId := fmt.Sprintf("%s:%s", rule.GetID(), "subject")
@@ -116,11 +111,11 @@ func (a *AuthorizerKetoEngineACPORY) Authorize(r *http.Request, session *authn.A
 	var b bytes.Buffer
 	u := fmt.Sprintf("%s://%s%s", r.URL.Scheme, r.URL.Host, r.URL.Path)
 
-	action, err := compiled.Replace(u, cf.RequiredAction, -1, -1)
+	action, err := rule.Replace(u, cf.RequiredAction)
 	if err != nil {
 		return errors.WithStack(err)
 	}
-	resource, err := compiled.Replace(u, cf.RequiredResource, -1, -1)
+	resource, err := rule.Replace(u, cf.RequiredResource)
 	if err != nil {
 		return errors.WithStack(err)
 	}
@@ -140,10 +135,10 @@ func (a *AuthorizerKetoEngineACPORY) Authorize(r *http.Request, session *authn.A
 	}
 
 	req, err := http.NewRequest("POST", urlx.AppendPaths(baseURL, "/engines/acp/ory", flavor, "/allowed").String(), &b)
-	req.Header.Add("Content-Type", "application/json")
 	if err != nil {
 		return errors.WithStack(err)
 	}
+	req.Header.Add("Content-Type", "application/json")
 
 	res, err := a.client.Do(req)
 	if err != nil {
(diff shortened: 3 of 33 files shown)


ayzu:
@aeneasr probably, it would make sense to discuss the draft, to verify whether I am heading the right direction.

aeneasr:
@ayzu yes, I will review it now. By the way, did you see the E-Mail from Jared?

ayzu:
Yes, I replied to it recently. Next week Thursday works for me, which
time works best for you?

driver/configuration/provider.go
```diff
@@ -19,6 +19,15 @@ const (
 	ForbiddenStrategyErrorType = "forbidden"
 )
 
+// MatchingStrategy defines matching strategy such as Regexp or Glob.
+type MatchingStrategy int
+
+// Possible values.
+const (
+	Regexp MatchingStrategy = iota
```
aeneasr: Would it maybe make more sense to have those as string constants `regexp` and `glob` - then it might be easier and read to debug if problems arise :)

rule/fetcher_default.go
```diff
@@ -197,12 +198,30 @@ func (f *FetcherDefault) watch(ctx context.Context, watcher *fsnotify.Watcher, e
...
 
 		return nil
 	})
 
-	f.enqueueEvent(events, event{et: eventRepositoryConfigChange, source: "entrypoint"})
+	var initialConfig map[string]interface{}
+
+	f.enqueueEvent(events, event{et: eventRepositoryConfigChanged, source: "entrypoint"})
+
+	viperx.AddWatcher(func(e fsnotify.Event) error {
```
aeneasr: Note to myself: Check if this will work as intended

rule/repository_memory.go
```diff
@@ -104,17 +121,18 @@ func (m *RepositoryMemory) Set(ctx context.Context, rules []Rule) error {
...
 	var rules []Rule
 	for k := range m.rules {
 		r := &m.rules[k]
-		if err := r.IsMatching(method, u); err == nil {
+		if matched, err := r.IsMatching(m.matchingStrategy, method, u); err != nil {
+			return nil, errors.WithStack(err)
+		} else if matched {
 			rules = append(rules, *r)
 		}
-		m.rules[k] = *r
```
aeneasr: This will disable caching of the matcher. I think that's ok for an interim solution but we should prevent recompiling the regular and glob expressions on every request. One thing to figure out is how that works depending on the matching strategy.

rule/rule.go
```diff
@@ -138,14 +140,6 @@ type Upstream struct {
 
 var _ json.Unmarshaler = new(Rule)
 
-func NewRule() *Rule {
```
aeneasr: A particular reason for removing this?

rule/rule.go
```diff
@@ -210,6 +209,17 @@ func (r *Rule) CompileURL() (*regexp.Regexp, error) {
...
 
+// Replace searches the input string and replaces each match (with the rule's pattern)
+// found with the replacement text.
+func (r *Rule) Replace(input, replacement string) (string, error) {
+	re, err := r.compileRegexp()
+	if err != nil {
+		return "", errors.WithStack(err)
+	}
+
+	return re.Replace(input, replacement, -1, -1)
```
aeneasr: Just a nitpick: while this is equivalent to `ReplaceAllString` I think that `ReplaceAllString` is more readable :)

rule/rule_test.go
```diff
@@ -42,7 +44,120 @@ func TestRule(t *testing.T) {
...
+		{
+			method:        "DELETE",
+			url:           "https://localhost/users/admin",
+			expectedMatch: false,
+			expectedErr:   nil,
+		},
+	}
+	for ind, tcase := range tests {
+		t.Run(string(ind), func(t *testing.T) {
+			matched, err := r.IsMatching(configuration.Regexp, tcase.method, mustParse(t, tcase.url))
```
aeneasr: I assume this will be `configuration.Glob` in the future? :)

rule/rule.go
ayzu: I didn't see usages of it.

pipeline/authz/keto_engine_acp_ory.go
```diff
@@ -115,9 +110,19 @@ func (a *AuthorizerKetoEngineACPORY) Authorize(r *http.Request, session *authn.A
 
 	var b bytes.Buffer
 	u := fmt.Sprintf("%s://%s%s", r.URL.Scheme, r.URL.Host, r.URL.Path)
+
+	action, err := rule.Replace(u, cf.RequiredAction)
```
ayzu: @aeneasr Do you intend to have this Replace functionality for regexp strategy only or for both regexp and glob?

aeneasr: Hm, that is a good question. I'm not sure if we can actually use this for glob, right?

I also think that we should really update this handler and make it more generic. So it would be fine for me to just support this specific handler for regexp, and throw an error if glob is used (for now).

What do you think?

ayzu: I agree. I don't see the out-of-the-box replacement functionality for glob.

We can add a check for a flavour here: https://github.com/ory/oathkeeper/blob/master/pipeline/authz/keto_engine_acp_ory.go#L111 and return an error if it is not a blank string (defaults to regex), "regex" or "text". Otherwise, we invoke `rule.ReplaceAllString` and proceed as usual. We can rename this method to `RegexpReplaceAll` to show that currently only regexp replacement is supported. What do you think? 

What's the difference, by the way, between "text" and "regex"? It seems like both flavours invoke `regexp.ReplaceAllString`.

ayzu: > Hm, that is a good question. I'm not sure if we can actually use this for glob, right?
> 
> I also think that we should really update this handler and make it more generic. So it would be fine for me to just support this specific handler for regexp, and throw an error if glob is used (for now).
> 
> What do you think?

I agree. I don't see the out-of-the-box replacement functionality for glob.

We can add a check for a flavour here: https://github.com/ory/oathkeeper/blob/master/pipeline/authz/keto_engine_acp_ory.go#L111 and return an error if it is not a blank string (defaults to regex), "regex" or "text". Otherwise, we invoke `rule.ReplaceAllString` and proceed as usual. We can rename this method to `RegexpReplaceAll` to show that currently only regexp replacement is supported. What do you think? 

What's the difference, by the way, between "text" and "regex"? It seems like both flavours invoke `regexp.ReplaceAllString`.

aeneasr: > return an error if it is not a blank string (defaults to regex), "regex" or "text". Otherwise, we invoke rule.ReplaceAllString and proceed as usual. We can rename this method to RegexpReplaceAll to show that currently only regexp replacement is supported. What do you think?
>
> What's the difference, by the way, between "text" and "regex"? It seems like both flavours invoke regexp.ReplaceAllString.

So the flavor is actually only relevant for the upstream service (ORY Keto), not for ORY Oathkeeper itself. Therefore, the flavor does not matter regarding the logic here - we can simply decide based on the access_rule matching strategy!

driver/configuration/provider.go
ayzu: yes, I agree. changed

pipeline/authz/keto_engine_acp_ory.go
ayzu: ok, thank you, @aeneasr 

As we discussed, I added a new interface `MatchingEngine` and its glob implementation. 

I will some tests and update documentation later.

rule/rule.go
```diff
@@ -49,9 +47,6 @@ type Match struct {
 	// You can use regular expressions in this field to match more than one url. Regular expressions are encapsulated in
```
ayzu: TODO: glob example

ayzu:
@aeneasr one more general question, about copyright in newly created files. Is it supposed to be the same as in existing ones? How about the copyright years?

aeneasr:
> @aeneasr one more general question, about copyright in newly created files. Is it supposed to be the same as in existing ones? How about the copyright years?

Yeah we haven't really been super diligent with that. We definitely need something to automate this in the future. Maybe via the CI or something.

ayzu:
@aeneasr I have created a corresponding PR in docs repo. I believe this PR now is review-ready:)

rule/matcher_test.go
```diff
@@ -119,3 +120,87 @@ func TestMatcher(t *testing.T) {
...
+		}
+	}
+
+	for name, matcher := range map[string]m{
+		"memory": NewRepositoryMemory(new(mockRepositoryRegistry)),
+	} {
+		t.Run(fmt.Sprintf("matcher=%s", name), func(t *testing.T) {
+			require.NoError(t, matcher.SetMatchingStrategy(context.Background(), configuration.Glob))
+			t.Run("case=empty", func(t *testing.T) {
+				testMatcher(t, matcher, "GET", "https://localhost:34/baz", true, nil)
```
ayzu: Not sure about merging this test, as we already have nested t.Run

ayzu:
Hi @aeneasr !

As discussed, I have merged the tests.

I have one check failing: https://circleci.com/gh/ory/oathkeeper/3381?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link

> [198/323] pkg:golang/github.com/opencontainers/runc@0.1.1  [Vulnerable]    1 known vulnerabilities affecting installed version

This is an indirect dependency. As far as I understand 0.1.1 is the latest stable version (though they do have some newer release candidates). 

What do you think?

aeneasr:
> This is an indirect dependency. As far as I understand 0.1.1 is the latest stable version (though they do have some newer release candidates).

The CVE affects everything below 1.0.0-rc.7 I believe. I've updated ory/x (which uses ory/dockertest, which uses runc) to bump runc -> dockertest -> x. If you update ory/x (which it is on master I believe so a merge/rebase should be enough) it should work :)

ayzu:
Yes, ory/x -> dockertst -> runc

ory/x v0.0.93 (latest tag) depends on runc 1.0.0-rc9 and we can use it.

Some other dependencies in `ory/oathkeeper` require the previous version of
ory/x:

 go mod graph | grep github.com/ory/x
github.com/ory/oathkeeper github.com/ory/x@v0.0.93
github.com/ory/sdk/swagutil@v0.0.0-20200202121523-307941feee4b
github.com/ory/x@v0.0.88
github.com/ory/go-acc@v0.1.0 github.com/ory/x@v0.0.85

Do you want me to update them? An alternative would be to use `replace`
statement in `ory/oathkeeper` for dependency `ory/x` 0.0.93

aeneasr:
Yeah, you have to update ory/x to 0.0.93 :) The library is very unstable, there might be build issues but they should be easy to resolve. If not, I can help

ayzu:
ok, @aeneasr 

I did a rebase, so now it uses ory/x 0.0.93

README.md
```diff
@@ -4,6 +4,8 @@
     <a href="https://discord.gg/PAMQWkr">Chat</a> |
     <a href="https://community.ory.sh/">Forums</a> |
     <a href="http://eepurl.com/di390P">Newsletter</a><br/><br/>
+    <a href="https://www.ory.sh/docs/guides/master/oathkeeper/">Guide</a> |
+    <a href="https://www.ory.sh/docs/oathkeeper/sdk/api">API Docs</a> |
```
aeneasr: These are duplicates

README.md
```diff
@@ -255,7 +257,6 @@ The Guide is available
 [here](https://www.ory.sh/docs/oathkeeper/).
 
 ### HTTP API documentation
-
```
aeneasr: Revert pls

rule/engine_glob.py.go
```diff
@@ -0,0 +1,95 @@
+package rule
```
aeneasr: A particular reason for naming this file `.py.go`? If not, let's rename it to `engine_glob.go` :)

rule/engine_regexp.go
```diff
@@ -0,0 +1,46 @@
+package rule
+
+import (
+	"hash/crc32"
```
aeneasr: Let's use `crc64` instead to reduce the likelihood of collisions :)

rule/engine_glob.py.go
```diff
@@ -0,0 +1,95 @@
...
+
+import (
+	"bytes"
+	"hash/crc32"
```
aeneasr: Let's use `crc64` instead to reduce the likelihood of collisions :)

rule/matcher_test.go
aeneasr: That's not a problem, we can nest `t.Run` I think merging the tests would save a bit of maintenance work later on :)

ayzu:
> Sweet! Just some minor things but I think this is good to merge after they have been addressed!

Thank you for the comments, I have updated the code.

aeneasr:
Awesome, thank you for your hard work!

ayzu:
Happy to commit and thank you for the reviews!

aeneasr:
No problem, will probably roll out a new release for this soon :)

Issue #346, opened 2020-01-22
High latencies


ecktom:
**Describe the bug**

Started using Oathkeeper and was quiet surprised about it's somewhat high latencies. I'm using a simple forward rule including (AllowAuthorizer / NoopAuthenticator / NoopMutator) 

```yaml
    - id: some-rule
      match:
        url: http://some-url/api/<.*>
        methods:
        - POST
        - GET
        - PATCH
        - PUT
      authenticators:
      - handler: noop
        config: {}
      authorizer:
        handler: allow
        config: {}
      mutators:
      - handler: noop
        config: {}
```

It takes around ~80ms for a request getting processed. I did some cpu profiling to check what happens. Turns out that https://github.com/ory/oathkeeper/blame/master/driver/configuration/provider_viper.go#L200 is getting called for every request multiplied with each authenticator, authorizer and mutator. This is calling the viper stack to compute the config which then get validated against the schema.

I just did some basic 3 line's caching there to proof my point and brought it down to ~5ms immediately (with all the schema validation still in place)

*CPU profile*

![pprof001](https://user-images.githubusercontent.com/3372410/72887202-83007000-3d0b-11ea-9bcf-2bda7133f894.png)

*Server configuration*
```yaml
    log:
      level: debug
      format: json
    profiling: cpu
    access_rules:
      repositories:
      - file:///etc/oathkeeper/rules.yaml
    authenticators:
      jwt:
        enabled: true
        config:
          jwks_urls:
          - https://some-url/jwks.json
      noop:
        enabled: true
        config: {}
    authorizers:
      allow:
        enabled: true
        config: {}
      deny:
        enabled: true
        config: {}
      keto_engine_acp_ory:
        enabled: true
        config:
          base_url: http://keto-url:4466/
          required_action: ""
          required_resource: ""
    mutators:
      noop:
        enabled: true
        config: {}
      header:
        enabled: true
        config:
          headers:
            X-SOME-HEADER: '{{ print .Subject }}'
```
**Expected behavior**

Would have expected some kind of caching here ;) I like the Prometheus approach where you would habe to call an api endpoint in order to reload the config.

**Environment**

* Version: v0.35.1
* Environment: Docker -> K8s

aeneasr:
Thank you for this amazing write-up! We've heard complaints in the recent days about slow performance when little CPU (e.g. 0.1 units) are available. This is most likely the root cause of that!

[...]

PR #348, opened 2020-01-24
Cache config to improve latencies

ecktom:
## Related issue
Fixes https://github.com/ory/oathkeeper/issues/346

Blocked by https://github.com/ory/viper/pull/7

## Proposed changes

Did a first draft here. Also had to hash the entire ENV in order to allow overrides which made things like 30-40% slower. ;) Any idea for that?

Guess we could tweak things even more eg. 
- pre-allocate + copy slices instead of append
- simple bit shift instead of binary.LittleEndian 

which however would make readability worse.

## Checklist


driver/configuration/provider_viper.go
```diff
@@ -197,7 +204,53 @@ func (v *ViperProvider) pipelineIsEnabled(prefix, id string) bool {
...
 func (v *ViperProvider) PipelineConfig(prefix, id string, override json.RawMessage, dest interface{}) error {
+	hash, err := v.hashPipelineConfig(prefix, id, override)
+	if err != nil {
+		return errors.WithStack(err)
+	}
+
+	c, ok := v.configCachge[hash]
+	if ok {
+		if dest != nil {
+			marshalled, err := json.Marshal(c)
```
aeneasr: Let's safe one round trip of encoding here - see line https://github.com/ory/oathkeeper/pull/348/files#diff-2449a6ea083767b149400840c05f41bdR308

I thought it might be a good idea to use glob, so I wrote a small benchmark. Unfortunately, glob seems to underperform when having to deal with interfaces (**so let's stick to json**), here are the results:

```
go test -bench . -benchmem
goos: darwin
goarch: amd64
pkg: github.com/ory/oathkeeper/docs
BenchmarkGobJSON/gob-12                     7848            151588 ns/op           55343 B/op       1234 allocs/op
BenchmarkGobJSON/json-12                   20253             57361 ns/op           32480 B/op        489 allocs/op
```

```go
package docs

import (
	"bytes"
	"encoding/gob"
	"encoding/json"
	"io/ioutil"
	"testing"

	"github.com/ghodss/yaml"
	"github.com/stretchr/testify/require"
)

func BenchmarkGobJSON(b *testing.B) {
	gob.Register(map[string]interface{}{})
	gob.Register([]interface{}{})

	c, err := ioutil.ReadFile(".oathkeeper.yaml")
	require.NoError(b, err)

	conf, err := yaml.YAMLToJSON(c)
	require.NoError(b, err)

	var o map[string]interface{}
	require.NoError(b, json.NewDecoder(bytes.NewBuffer(conf)).Decode(&o))

	var gb bytes.Buffer
	require.NoError(b, gob.NewEncoder(&gb).Encode(&o))

	b.Run("gob", func(b *testing.B) {
		for i := 0; i < b.N; i++ {
			b.StopTimer()
			m := make(map[string]interface{})
			buf := bytes.NewBuffer(gb.Bytes())
			b.StartTimer()

			require.NoError(b, gob.NewDecoder(buf).Decode(&m))
		}
	})

	b.Run("json", func(b *testing.B) {
		for i := 0; i < b.N; i++ {
			b.StopTimer()
			m := make(map[string]interface{})
			buf := bytes.NewBuffer(conf)
			b.StartTimer()

			require.NoError(b, json.NewDecoder(buf).Decode(&m))
		}
	})
}
```

driver/configuration/provider_viper.go
```diff
@@ -305,7 +299,13 @@ func (v *ViperProvider) PipelineConfig(prefix, id string, override json.RawMessa
 		return errors.WithStack(result.Errors())
 	}
 
-	v.configCachge[hash] = config
+	marshalled, err = json.Marshal(config)
```
aeneasr: `marshalled` should already be the proper type and payload (see https://github.com/ory/oathkeeper/pull/348/files/ede19d9b718fa4d91191ae9d138e89546305440c..8231b9aa87680589b8c42c25a46c5188acc446ca#diff-2449a6ea083767b149400840c05f41bdR296) so I don't think we need to encode it again!

driver/configuration/provider_viper_public_test.go
```diff
@@ -48,11 +48,9 @@ func TestPipelineConfig(t *testing.T) {
 		require.NoError(t, p.PipelineConfig("authenticators", "oauth2_introspection", nil, &res))
 		assert.JSONEq(t, `{"introspection_request_headers":{},"introspection_url":"https://override/path","pre_authorization":{"client_id":"some_id","client_secret":"some_secret","enabled":true,"scope":["foo","bar"],"token_url":"https://my-website.com/oauth2/token"},"required_scope":[],"scope_strategy":"exact","target_audience":[],"trusted_issuers":[]}`, string(res), "%s", res)
 
+		// Cleanup
 		require.NoError(t, os.Setenv("AUTHENTICATORS_OAUTH2_INTROSPECTION_CONFIG_INTROSPECTION_URL", ""))
 
-		require.NoError(t, p.PipelineConfig("authenticators", "oauth2_introspection", nil, &res))
```
aeneasr: Why were these removed?

driver/configuration/provider_viper_public_test.go
```diff
@@ -93,6 +91,38 @@ func TestPipelineConfig(t *testing.T) {
...
 }
 
+/*
+go test ./... -v -bench=. -run BenchmarkPipelineConfig -benchtime=10s
+
+v0.35.1
+594	  20119202 ns/op
+
+v0.35.2
+3048037	      3908 ns/op
```
aeneasr: Wow, that perf improvement!

driver/configuration/provider_viper_public_test.go
ecktom: This was basically the root cause of my intention to cache the environment ;) But as you've correctly pointed out there will be no changes to the env for the running process... So I've simply adjusted the test to reflect this.

driver/configuration/provider_viper.go
ecktom: yeah :D quiet obviously... Removed the unnecessary marshall...

driver/configuration/provider_viper_public_test.go
ecktom: Indeed :D

driver/configuration/provider_viper_public_test.go
aeneasr: Ohh, I see - that totally makes sense :)

aeneasr:
Awesome, thank you so much for your hard work!

PR #353, opened 2020-01-26
Hash enabled check to further improve performance

ecktom:
## Related issue
https://github.com/ory/oathkeeper/issues/346

## Proposed changes

Just found one more thing to cache ;) Also thought about combining it with PipelineConfig but thought this one is clearer to understand and also with less changes

## Checklist

## Further comments
Brings it down from 2 - 3ms to 0.1 - 0.3ms

Before:
profile002

After:
profile001

### DIFF
--- a/driver/configuration/provider_viper.go
+++ b/driver/configuration/provider_viper.go
@@ -104,15 +104,20 @@ const (
 )
 
 type ViperProvider struct {
-	l            logrus.FieldLogger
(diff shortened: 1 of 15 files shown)


aeneasr:
This is pretty insane! I would not have thought that viper has such a slow retrieval process! I think we should probably introduce caching to viper itself because this really isn't acceptable performance-wise!

Thank you for this change :)

PR #379, opened 2020-03-17
fix: Don't send scope with OAuth2 introspection request

### LINKED ISSUE #377: Hierarchic scope strategy not working for OAuth2 Introspection authn
**Describe the bug**

I'm running into an issue authenticating a request with oathkeeper using the hierarchic scope strategy. 
I'm trying to lock down my hydra introspection endpoint.

I have added the following to the oathkeeper config to enable the oauth2 introspection authenticator
```
authenticators:
  oauth2_introspection:
    enabled: true
    config:
      introspection_url: http://localhost:4445/oauth2/introspect
      scope_strategy: hierarchic
      introspection_request_headers:
        x-forwarded-proto: https
```

I have added the following rule to protect my hydra introspection endpoint
```
- id: admin:introspect-access-token
  upstream:
    url: http://localhost:4445
    preserve_host: true
    strip_path: /v1/op
  match:
    url: http://<[a-zA-Z0-9-_:\.]+>/v1/op/oauth2/introspect
    methods:
      - POST
  authenticators:
    - handler: oauth2_introspection
      config:
        required_scope:
          - admin.introspect
  authorizer:
    handler: allow
  mutators:
    - handler: noop
```

I have created the following client for administrative purposes
```
{
    "client_id": "090e30ed-bc4b-478d-bb82-975a1667334b",
    "client_name": "Admin",
    "redirect_uris": [],
    "grant_types": [
        "client_credentials"
    ],
    "response_types": [],
    "scope": "admin",
    "audience": [],
    "owner": "",
    "policy_uri": "",
    "allowed_cors_origins": [],
    "tos_uri": "",
    "client_uri": "",
    "logo_uri": "",
    "contacts": [],
    "client_secret_expires_at": 0,
    "subject_type": "public",
    "token_endpoint_auth_method": "client_secret_basic",
    "userinfo_signed_response_alg": "none",
    "created_at": "2020-03-10T13:43:55Z",
    "updated_at": "2020-03-10T13:43:55Z",
    "metadata": null
}
```

Using that client I have obtained the following token.
```
VRS1H90d3XeP5xX18EeiIG31RHvU-1ewUzcGRLHwT1U.DBqrTNGO0d5h4s8Jn6-3sBGsYvQsL87hjy6PvZDD3Co
```

I then attempted to introspect that token with the following request (Format from insomnia timeline)
```
> POST /v1/op/oauth2/introspect HTTP/2
> Host: myapp.com
> User-Agent: insomnia/7.1.1
> Content-Type: application/x-www-form-urlencoded
> Authorization: Bearer VRS1H90d3XeP5xX18EeiIG31RHvU-1ewUzcGRLHwT1U.DBqrTNGO0d5h4s8Jn6-3sBGsYvQsL87hjy6PvZDD3Co
> Accept: */*
> Content-Length: 93

| token=VRS1H90d3XeP5xX18EeiIG31RHvU-1ewUzcGRLHwT1U.DBqrTNGO0d5h4s8Jn6-3sBGsYvQsL87hjy6PvZDD3Co
```

oathkeeper, which is protecting the introspection endpoint, responded with the following
```
{
  "error": {
    "code": 401,
    "status": "Unauthorized",
    "request": "bec3fc5f-2569-92ed-8ae7-3925cc3677a0",
    "reason": "Access token i says token is not active",
    "message": "Access credentials are invalid"
  }
}
```

I then inspected the same token again but this time using a different Authorization token which has the scope of `admin.introspect`.
```
> POST /v1/op/oauth2/introspect HTTP/2
> Host: myapp.com
> User-Agent: insomnia/7.1.1
> Content-Type: application/x-www-form-urlencoded
> Authorization: Bearer XPx_x65ReH10Di-HALA62l9TYAQJPZth4t7e7apYMSc.jHVDuDLOqyt9sd6gHQ5Gw2WYlCq7oAk0CBLkbaTgX9M
> Accept: */*
> Content-Length: 93

| token=VRS1H90d3XeP5xX18EeiIG31RHvU-1ewUzcGRLHwT1U.DBqrTNGO0d5h4s8Jn6-3sBGsYvQsL87hjy6PvZDD3Co
```

This time the request to was successful and I received the following response.
```
{
  "active": true,
  "scope": "admin",
  "client_id": "090e30ed-bc4b-478d-bb82-975a1667334b",
  "sub": "090e30ed-bc4b-478d-bb82-975a1667334b",
  "exp": 1583860409,
  "iat": 1583856809,
  "iss": "https://myapp.com/v1/op/",
  "token_type": "access_token"
}
```

As you can see the introspected token is active and has the `admin` scope. The `admin` scope *should* match `admin.introspect` using the hierarchic scope strategy.

I investigated a little further and found that oathkeeper is sending the required_scope to hydra when inspecting the token.
https://github.com/ory/oathkeeper/blob/master/pipeline/authn/authenticator_oauth2_introspection.go#L88

Since `admin.introspect`, the required scope, doesn't equal `admin`, the granted scope, hydra responds with the following when introspecting the token
```
{
  "active": false
}
```

Oathkeeper sees this response and rejects the request.

It seems the me that the solution is for oathkeeper to not include the scope when introspecting the token and instead validate the scope itself (which it already does further down)
https://github.com/ory/oathkeeper/blob/master/pipeline/authn/authenticator_oauth2_introspection.go#L132-L138

**Environment**

* Version: 
- Oathkeeper: oryd/oathkeeper:v0.33.0-beta.1
- Hydra: oryd/hydra:v1.1.0-alpine
* Environment: Docker

tleef:
## Related issue

Fixes #377 

## Proposed changes

Oathkeeper shouldn't send scope with introspection requests and instead should rely on its own scope validation. See linked issue for more details.

## Checklist

- [x] I have documented my changes in the [developer guide](https://github.com/ory/docs) (if appropriate)

### DIFF
--- a/pipeline/authn/authenticator_oauth2_introspection.go
+++ b/pipeline/authn/authenticator_oauth2_introspection.go
@@ -85,7 +85,7 @@ func (a *AuthenticatorOAuth2Introspection) Authenticate(r *http.Request, session
 		return errors.WithStack(ErrAuthenticatorNotResponsible)
 	}
 
-	body := url.Values{"token": {token}, "scope": {strings.Join(cf.Scopes, " ")}}
+	body := url.Values{"token": {token}}
 	introspectReq, err := http.NewRequest(http.MethodPost, cf.IntrospectionURL, strings.NewReader(body.Encode()))
 	if err != nil {
 		return errors.WithStack(err)
--- a/pipeline/authn/authenticator_oauth2_introspection.go
+++ b/pipeline/authn/authenticator_oauth2_introspection.go
@@ -86,6 +86,12 @@ func (a *AuthenticatorOAuth2Introspection) Authenticate(r *http.Request, session
 	}
 
 	body := url.Values{"token": {token}}
+
+	ss := a.c.ToScopeStrategy(cf.ScopeStrategy, "authenticators.oauth2_introspection.scope_strategy")
+	if ss == nil {
+		body.Add("scope", strings.Join(cf.Scopes, " "))
+	}
+
 	introspectReq, err := http.NewRequest(http.MethodPost, cf.IntrospectionURL, strings.NewReader(body.Encode()))
 	if err != nil {
 		return errors.WithStack(err)
@@ -129,7 +135,7 @@ func (a *AuthenticatorOAuth2Introspection) Authenticate(r *http.Request, session
 		}
 	}
 
-	if ss := a.c.ToScopeStrategy(cf.ScopeStrategy, "authenticators.oauth2_introspection.scope_strategy"); ss != nil {
+	if ss != nil {
 		for _, scope := range cf.Scopes {
 			if !ss(strings.Split(i.Scope, " "), scope) {
 				return errors.WithStack(helper.ErrForbidden.WithReason(fmt.Sprintf("Scope %s was not granted", scope)))
--- a/pipeline/authn/authenticator_oauth2_introspection_test.go
+++ b/pipeline/authn/authenticator_oauth2_introspection_test.go
@@ -70,7 +70,7 @@ func TestAuthenticatorOAuth2Introspection(t *testing.T) {
 					m.POST("/oauth2/introspect", func(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
 						require.NoError(t, r.ParseForm())
 						require.Equal(t, "token", r.Form.Get("token"))
-						require.Equal(t, "scope-a scope-b", r.Form.Get("scope"))
+						require.Equal(t, "", r.Form.Get("scope"))
 						w.WriteHeader(http.StatusNotFound)
 					})
 				},
@@ -205,7 +205,7 @@ func TestAuthenticatorOAuth2Introspection(t *testing.T) {
 					m.POST("/oauth2/introspect", func(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
 						require.NoError(t, r.ParseForm())
 						require.Equal(t, "token", r.Form.Get("token"))
-						require.Equal(t, "scope-a scope-b", r.Form.Get("scope"))
+						require.Equal(t, "", r.Form.Get("scope"))
 						require.NoError(t, json.NewEncoder(w).Encode(&AuthenticatorOAuth2IntrospectionResult{
 							Active:   true,
 							Subject:  "subject",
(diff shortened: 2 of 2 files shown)


pipeline/authn/authenticator_oauth2_introspection.go
```diff
@@ -85,7 +85,7 @@ func (a *AuthenticatorOAuth2Introspection) Authenticate(r *http.Request, session
 		return errors.WithStack(ErrAuthenticatorNotResponsible)
 	}
 
-	body := url.Values{"token": {token}, "scope": {strings.Join(cf.Scopes, " ")}}
```
aeneasr: Exclude this only when the scope strategy is set to `none` :)

tleef: I think we exclude it in all cases. 

For example, if the scope strategy of the given Oathkeeper rule is `wildcard`, Hydra won't know that and therefore won't be able to properly validate the scope which is why we opt out of Hydra's scope validation all together and rely only on Oathkeeper's validation which happens after the token introspection. 

https://github.com/ory/oathkeeper/blob/master/pipeline/authn/authenticator_oauth2_introspection.go#L134

That logic applies no matter what the scope strategy for a given Oathkeeper rule is.

When the scope strategy is `none`, that means we still want to opt out of Hydra's validation but we also want to opt out of Oathkeeper's validation in favor of the application's scope validation. This is useful for GraphQL APIs where it's not possible to know what scopes are required based solely on the path name.

The basic problem is Hydra can't adapt its behavior bases on the scope strategy. For that to be possible, Hydra would need to accept the strategy as another parameter to the introspection endpoint. Also, Oathkeepr doesn't know the scope strategy that Hydra is configured to use so it can't conditionally send the scope to Hydra based on the scope strategy.

aeneasr: I disagree with your sentiment. The scope is enforced by the Authorization Server and only in cases where we explicitly want to override this behavior we use a specific strategy for Hydra.

Making this change now will leave unintended breaking changes for all people that rely on this feature for authorization. It is thus not a good idea to completely remove it.

> When the scope strategy is none, that means we still want to opt out of Hydra's validation but we also want to opt out of Oathkeeper's validation in favor of the application's scope validation. This is useful for GraphQL APIs where it's not possible to know what scopes are required based solely on the path name.

If you want to opt out of any validation just don't include any scopes in the rule.

tleef: Okay, that makes sense. So we *only* send the scope to Hydra when the scope strategy is `none`. Let me make that change

aeneasr: Awesome, thank you! :)

tleef: I added some tests for the different strategies. Before, all of the tests where running with the `exact` strategy. Now there are at least two tests for each. One that passes and one that fails. And the tests default to a strategy of `none` which is in line with the default behavior of Oathkeeper.

aeneasr: Awesome!

Issue #387, opened 2020-03-25
Unauthorized instead of Forbidden status with client_credentials


colunira:
**Describe the bug**

I have two Rules pointing to one service which require different scopes for different endpoints (read or write). When I try to reach endpoint A, which requires "read" scope, with a token with "read" scope I get a correct response. However, when I try to reach endpoint B, which requires "write" scope, with the same token with "read" scope (which I know is proper, because I can connect to service A with it), I get 401 Unauthorized instead of expected 403 Forbidden. 

**Reproducing the bug**

1. Create a Rule and a Oauth2Client:
```
apiVersion: oathkeeper.ory.sh/v1alpha1
kind: Rule
metadata:
  name: httpbin-8m82z
  namespace: default
    kind: APIRule
    name: httpbin
spec:
  authenticators:
  - config:
      required_scope:
      - write
    handler: oauth2_introspection
  authorizer:
    handler: allow
  match:
    methods:
    - POST
    url: <http|https>://httpbin.kyma.local</post>
  upstream:
    url: http://httpbin.default.svc.cluster.local:8000
```
```
apiVersion: hydra.ory.sh/v1alpha1
kind: OAuth2Client
metadata:
  name: test-client
  namespace: default
spec:
  grantTypes:
  - "client_credentials"
  scope: "read write"
  secretName: test-secret
```

2. Fetch the access token from Hydra with scope read using data from connected to OAuth2Client secret:

```curl -X POST "https://{ADDRESS}/oauth2/token" -H "Authorization: Basic {CREDENTIALS}" -F "grant_type=client_credentials" -F "scope=write"```

3. Call exposed service:

```curl -X POST "https://httpbin.kyma.local/post" -H "Authorization: Bearer {ACCESS_TOKEN}"```

*Oathkeeper logs*

```
time="2020-03-25T11:30:39Z" level=info msg="started handling request" method=GET remote="127.0.0.1:49626" request=/headers request_id=48e57029-c9cc-4a45-9564-6afc6e0f8bef
time="2020-03-25T11:30:39Z" level=warning msg="The authentication handler encountered an error" authentication_handler=oauth2_introspection error="Access credentials are invalid" granted=false http_host=httpbin.kyma.local http_method=GET http_url="http://httpbin.kyma.local/headers" http_user_agent=curl/7.66.0 reason_id=authentication_handler_error rule_id=httpbin-kh4vn.default
[...]

PR #411, opened 2020-04-16
fix: return a 403 when introspection result is inactive

flaviuvadan:
## Related issue
The issue has been reported 22 days ago, today, and can be found [here](https://github.com/ory/oathkeeper/issues/387).

## Proposed changes
Make the active check return a 403 instead of a 401.

## Checklist
      the changes.
      works.

## Further comments
The issue describes how a 401 is returned instead of an expected 403 upon a scope check failure. The 401 is returned by the condition that checks whether the introspection result is active. The author of the issue describes how it was expected for `Authenticate` to return a 403 upon a scope check failure. However, based on what the function is doing, it seems that the introspection result was listed as inactive, which caused the 401 to be returned. I am not aware of whether the `!i.Active` block returns a 401 by design.

### DIFF
--- a/pipeline/authn/authenticator_oauth2_introspection.go
+++ b/pipeline/authn/authenticator_oauth2_introspection.go
@@ -120,7 +120,7 @@ func (a *AuthenticatorOAuth2Introspection) Authenticate(r *http.Request, session
 	}
 
 	if !i.Active {
-		return errors.WithStack(helper.ErrUnauthorized.WithReason("Access token i says token is not active"))
+		return errors.WithStack(helper.ErrForbidden.WithReason("Access token i says token is not active"))
 	}
 
 	for _, audience := range cf.Audience {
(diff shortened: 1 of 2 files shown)


Marlinc:
I'm not sure I feel comfortable returning a 403 Forbidden when a access token has expired. Sure it makes sense for missing scopes. So maybe we should extend Hydra so it can return whether or not the token has expired or not. In that case we return a 401 Unauthorized and then return 403 Forbidden when the token is otherwise 'inactive`

aeneasr:
Yes, an invalid token is a 401, not a 403, regardless of why the uptream auth server thinks. We could return a 403 when the scope strategy of oathkeeper mismatches.

aeneasr:
I won't be able to merge the PR as is though because the existing behavior is expected. Will the changes be made or should I close this PR?

flaviuvadan:
Alright, thanks for the feedback! Let's close it. Based on the feedback, I may have looked in the wrong place. I will explore the codebase more to see how the scope strategy match is performed and put up a different PR.

Issue #417, opened 2020-04-23
Cache hydrate calls


pike1212:
**Is your feature request related to a problem? Please describe.**

Would like to reduce latency by caching hydrate calls

**Describe the solution you'd like**

Cache calls similar to the id token caching

**Additional context**

I've already started on a PR

aeneasr:
Nice!

PR #418, opened 2020-04-24
feat: add cache to hydrator

pike1212:
## Related issue

#417 

## Proposed changes

Add ability to configure a cache for hydrate calls

## Checklist

      the changes.
      works.

## Further comments

Works similar to the id token cache, but uses a `cache_ttl` config to determine if it should cache calls

### DIFF
--- a/.schema/config.schema.json
+++ b/.schema/config.schema.json
@@ -843,6 +843,12 @@
               "$ref": "#/definitions/retry"
             }
           }
+        },
+        "cache_ttl": {
+          "type": "string",
+          "pattern": "^[0-9]+(ns|us|ms|s|m|h)$",
+          "title": "Cache Time to Live",
+          "description": "How long to cache hydrate calls"
         }
       },
       "required": [
--- a/.schema/config.schema.json
+++ b/.schema/config.schema.json
@@ -844,11 +844,20 @@
             }
           }
         },
-        "cache_ttl": {
-          "type": "string",
-          "pattern": "^[0-9]+(ns|us|ms|s|m|h)$",
-          "title": "Cache Time to Live",
-          "description": "How long to cache hydrate calls"
+        "cache": {
+          "additionalProperties": false,
+          "required": [
+            "ttl"
+          ],
+          "type": "object",
+          "properties": {
+            "ttl": {
+              "type": "string",
+              "pattern": "^[0-9]+(ns|us|ms|s|m|h)$",
+              "title": "Cache Time to Live",
+              "description": "How long to cache hydrate calls"
+            }
+          }
         }
       },
       "required": [
--- a/pipeline/mutate/mutator_hydrator.go
+++ b/pipeline/mutate/mutator_hydrator.go
@@ -60,8 +60,7 @@ type MutatorHydrator struct {
 	client *http.Client
 	d      mutatorHydratorDependencies
 
-	hydrateCache        *ristretto.Cache
-	hydrateCacheEnabled bool
+	hydrateCache *ristretto.Cache
 }
 
 type BasicAuth struct {
@@ -85,15 +84,12 @@ type externalAPIConfig struct {
 }
 
 type MutatorHydratorConfig struct {
-	Api externalAPIConfig `json:"api"`
+	Api      externalAPIConfig `json:"api"`
+	CacheTtl string            `json:"cache_ttl"`
 }
 
 type mutatorHydratorDependencies interface {
 	x.RegistryLogger
-}
-
-func (a *MutatorHydrator) SetCaching(cache bool) {
-	a.hydrateCacheEnabled = cache
 }
 
 func NewMutatorHydrator(c configuration.Provider, d mutatorHydratorDependencies) *MutatorHydrator {
@@ -102,7 +98,7 @@ func NewMutatorHydrator(c configuration.Provider, d mutatorHydratorDependencies)
 		MaxCost:     1 << 25,
 		BufferItems: 64,
 	})
-	return &MutatorHydrator{c: c, d: d, client: httpx.NewResilientClientLatencyToleranceSmall(nil), hydrateCache: cache, hydrateCacheEnabled: true}
+	return &MutatorHydrator{c: c, d: d, client: httpx.NewResilientClientLatencyToleranceSmall(nil), hydrateCache: cache}
 }
 
 func (a *MutatorHydrator) GetID() string {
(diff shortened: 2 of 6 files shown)


pipeline/mutate/mutator_hydrator.go
```diff
@@ -78,21 +84,53 @@ type externalAPIConfig struct {
...
 
 type mutatorHydratorDependencies interface {
 	x.RegistryLogger
 }
 
 func NewMutatorHydrator(c configuration.Provider, d mutatorHydratorDependencies) *MutatorHydrator {
-	return &MutatorHydrator{c: c, d: d, client: httpx.NewResilientClientLatencyToleranceSmall(nil)}
+	cache, _ := ristretto.NewCache(&ristretto.Config{
+		NumCounters: 10000,
+		MaxCost:     1 << 25,
```
aeneasr, line 98: How much RAM will this use at maximum use? Should this be configurable?

pipeline/mutate/mutator_hydrator.go
```diff
@@ -78,21 +84,53 @@ type externalAPIConfig struct {
...
+	return &MutatorHydrator{c: c, d: d, client: httpx.NewResilientClientLatencyToleranceSmall(nil), hydrateCache: cache}
 }
 
 func (a *MutatorHydrator) GetID() string {
 	return "hydrator"
 }
 
+func (a *MutatorHydrator) cacheKey(config *MutatorHydratorConfig, session *authn.AuthenticationSession) string {
+	return fmt.Sprintf("%x",
+		md5.Sum([]byte(fmt.Sprintf("%s|%s", config.Api.URL, session.Subject))),
```
aeneasr, line 110: Since the string is not that long I think we can use it without md5 as it saves a couple of CPU rounds :)

.schema/config.schema.json
```diff
@@ -843,6 +843,12 @@
               "$ref": "#/definitions/retry"
             }
           }
+        },
+        "cache_ttl": {
```
aeneasr, line 847: I'd like to propose another structure:

```
cache:
  ttl:
```

This would allow us to add more keys later on:

```
cache:
  strategy: ristretto # groupcache? lru?
  size: ..
  ...
```

pipeline/mutate/mutator_hydrator.go
pike1212: I think it should be configurable, not sure the best way to do that as the cache is created when it is initialized, but the config comes through on the hydrate method, any suggestions?

aeneasr: Yeah I think that makes sense but we can also just leave it as is. I think we can add config later if it's really required!

pipeline/mutate/mutator_hydrator.go
```diff
@@ -77,22 +82,56 @@ type externalAPIConfig struct {
...
 type mutatorHydratorDependencies interface {
 	x.RegistryLogger
 }
 
 func NewMutatorHydrator(c configuration.Provider, d mutatorHydratorDependencies) *MutatorHydrator {
-	return &MutatorHydrator{c: c, d: d, client: httpx.NewResilientClientLatencyToleranceSmall(nil)}
+	cache, _ := ristretto.NewCache(&ristretto.Config{
+		NumCounters: 10000,
+		MaxCost:     1 << 25,
+		BufferItems: 64,
```
aeneasr, line 102: I think it makes sense to annotate the config a bit:

```suggestion
		// This will hold about 1000 unique mutation responses.
		NumCounters: 10000,
		// Allocate a max of 32MB
		MaxCost:     1 << 25,
		// This is a best-practice value.
		BufferItems: 64,
```

pipeline/mutate/mutator_hydrator.go
```diff
@@ -104,6 +143,13 @@ func (a *MutatorHydrator) Mutate(r *http.Request, session *authn.AuthenticationS
 		return errors.WithStack(err)
 	}
 
+	if cfg.Cache.Ttl != "" {
```
aeneasr, line 146: Should we not also respsect `Cache-Control` headers? Maybe the server doesn't want us to cache the requests and sends `Cache-Control: no-cache` or some other cache value?

aeneasr: There's actually a http.NetTransport that does that: https://github.com/gregjones/httpcache

pipeline/mutate/mutator_hydrator.go
```diff
@@ -77,22 +82,56 @@ type externalAPIConfig struct {
 	Retry *retryConfig `json:"retry"`
 }
 
+type cacheConfig struct {
+	Ttl string `json:"ttl"`
```
aeneasr, line 86:
```suggestion
	TTL string `json:"ttl"`
```

pipeline/mutate/mutator_hydrator.go
```diff
@@ -174,6 +220,15 @@ func (a *MutatorHydrator) Mutate(r *http.Request, session *authn.AuthenticationS
 	}
 	*session = sessionFromUpstream
 
+	if cfg.Cache.Ttl != "" {
```
aeneasr, line 223: Let's make this a private field `ttl time.Duration` and parse it in `func (a *MutatorHydrator) Config(config json.RawMessage) (*MutatorHydratorConfig, error)` :)

pipeline/mutate/mutator_hydrator.go
pike1212: That library will only work for GET requests. I imagine the hydrator is mostly used where someone is controlling both Oathkeeper and the hydrate service.

aeneasr: Ah, I see - yeah I think it makes sense. We can also add this at a later stage anyways as it won't be a breaking change.

aeneasr:
Linter complains because of format: `make format; git commit -a -m "format"; git push` :)

aeneasr:
Thank you, this looks perfect :)

pike1212:
I found an issue where the downstream id token mutator adds header information that gets put into the hydrate cache, and with a large volume of calls will sometimes cause a concurrent write error in the id token mutator.

aeneasr:
Do you know how to fix this?

pike1212:
I was able to fix it by using a deep copy library in and out of the cache (https://github.com/mitchellh/copystructure). The other option might be to just cache the json response instead of the de-serialized struct.

PR #422, opened 2020-04-27
Don't copy the decission endpoint request's Content-Length

Marlinc:
We currently copy all original request headers send to the decission
endpoint back. This can include the Content-Length header which
describes the request body or response. Including the original
request Content-Length causes issues for the decission endpoint
client if the response body doesn't match the exact size.

## Proposed changes

This change makes sure the Content-Length doesn't get included in
the response body and adds a test to prevent future regressions.

## Checklist

      the changes.
      works.

### DIFF
--- a/api/decision.go
+++ b/api/decision.go
@@ -122,6 +122,11 @@ func (h *DecisionHandler) decisions(w http.ResponseWriter, r *http.Request) {
 		Info("Access request granted")
 
 	for k := range s.Header {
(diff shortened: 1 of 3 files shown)


aeneasr:
Thanks!

flusflas:
Is this issue really solved? I have the same problem: the "Content-Length" of the request and the response are the same, even when the response has an empty body.

PR #416, opened 2020-04-20
Add new remote authorizer that uses request body and headers

Marlinc:
## Related issue

None that I know

## Proposed changes

This pull request implements a new authorizer that sends the original request body as body to the remote endpoint. This allows the remote endpoint to take the body into account in its decision.

## Checklist

- [x] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md).
- [x] I have read the [security policy](../security/policy).
- [x] I confirm that this pull request does not address a security
      vulnerability. If this pull request addresses a security. vulnerability, I
      confirm that I got green light (please contact
      [security@ory.sh](mailto:security@ory.sh)) from the maintainers to push
      the changes.
- [x] I have added tests that prove my fix is effective or that my feature
      works.
- [x] I have added or changed [the documentation](docs/docs).

PR #426, opened 2020-04-30
fix: use pipe to pass body remote authorizer

Marlinc:
## Proposed changes

The big request test introduced in #416 appears to be flaky and fails every so often. I've made some changes to it and have also changed the piping to use a pipe to pass the original request to the `remote` authorizer. Instead of creating two in memory buffers.

## Checklist

      the changes.
      works.


aeneasr:
Thank you!

Issue #98, opened 2018-08-08
Add OPA authorizer


aeneasr:
https://github.com/open-policy-agent/opa

aeneasr:
It's not clear how to work with OPA's unstructured data approach. I'm closing this until there is clarification. Feel free to discuss this though.

tloriato:
How did this progress?

zepatrik:
Keto as a separate project was started. We are currently in the process of rewriting it from the ground up, a lot of exciting stuff happening there :wink: OPA integration will be possible there still, see https://github.com/ory/keto/issues/318

Issue #441, opened 2020-05-18
 Lessons learned & ORY Oathkeeper NextGen


aeneasr:
First of all, thank you for using ORY Oathkeeper and placing trust in us! This issue's purpose is to discuss structural changes to ORY Oathkeeper.

ORY Oathkeeper is used in production at many companies, handling tons of traffic every day. As such, we naturally receive a lot of bug reports and pull requests. Let's take a look at areas of improvement. Some of these changes will not be backwards compatible, which is why we want you to discuss them with us!

All of the listed changes are breaking changes, which is why it is important to get your feedback! When giving feedback, please keep in mind that we want to build the best product possible for you! This sometimes includes breaking things to move forward :)

## Non-Goals

Before we start I want to define some non-goals. One of the non-goals is to implement an "identity-aware" proxy like [Pomerium](https://www.pomerium.io). The reason for that is that we don't want to have user interfaces in Oathkeeper itself. It will be possible (in future versions) to get what Pomerium does (plus a ton more) by combining ORY Oathkeeper and ORY Kratos. If an identity-aware proxy is all that's needed, Pomerium is also probably the better choice!

## Support negative lookahead

Neither glob nor regex currently support negative lookahead, but it is quite important to support them as it allows for much easier rule definition. See also https://www.linuxjournal.com/content/bash-extended-globbing

## More Proxies

https://github.com/ory/oathkeeper/pull/265

## Logging
This release aims to better support debugging by adding a lot of debug logs. We understand that debugging ORY Oathkeeper is time-consuming and frustrating and we want to provide the best tooling possible to fix that!

**This is a non-breaking change.**

## Error Handling

The error handling is quite complex to understand, configure, and debug. The following config is a real-world example:

```
errors:
  fallback:
    - json

  handlers:
    redirect:
      enabled: true
      config:
        to: http://127.0.0.1:4455/auth/login
        when:
          -
            error:
              - unauthorized
              - forbidden
            request:
              header:
                accept:
                  - text/html
    json:
      enabled: true
      config:
        verbose: true
```

All it does is return JSON unless the a real user using a browser is making the request, in which case we redirect. We need something better!

These change will not be backwards compatible.

### Redirect

One thing we can improve is to decide what to do based on the accept header. If it includes `application/json`, we return JSON (what else would we return?). If it includes
`text/html` we show an HTML error.

To redirect the browser on error, we could define something like:

```
errors:
  browser:
    when:
      unauthorized:
[...]

PR #462, opened 2020-06-05
Add remote_opa authorizer

ryanph:
## Related issue

Issue #98 requests an OPA authoriser

## Proposed changes

Implement an OPA authorizer that queries an OPA server via its REST API for authorization. OPA can also be used as a Go library directly within applications and as this authorizer interfaces with an OPA server using the REST API it is called 'remote_opa' rather than just 'opa'.

The existing generic http authorizers (remote and remote_json) rely on the remote service responding with a 200 or 403 response code to authorize requests. When using the OPA REST API to query a policy document with input (e.g. the details of a HTTP request) the response code is always 200 and it is the response body that contains the result of the policy. As such the existing authorizers are not suitable. 

A quick way to test / demonstrate this authorizer:

Start an OPA instance:
`docker run -p 8181:8181 openpolicyagent/opa run --server`

Upload a basic policy:
```
cat <<EOF > policy.rego
package example.authz

import input

default allow = false

allow = true {
    input.method = "GET"
    input.path = ["profiles"]
}

allow = true {
    input.method = "POST"
    input.path = ["profiles", username ]
    input.user == username
}
EOF
```

`curl -X PUT --data-binary @example/policy.rego localhost:8181/v1/policies/example`

In this example policy anyone can GET /profiles, and users can POST to their own profile under /profiles/username (or in this example with the anonymous authenticator anyone can POST to /profiles/anonymous).

Configure the remote_opa authorizer to use this policy:
```
cat <<EOF > oathkeeper.yml
serve:
  proxy:
    port: 8000

authenticators:
  noop:
    enabled: true
  anonymous:
    enabled: true

authorizers:
  remote_opa:
    enabled: true
    config:
      remote: http://127.0.0.1:8181/v1/data/example/authz 
      payload: "{\"input\":{\"user\":\"{{ .Subject }}\",\"path\":{{ .PathArray }},\"method\":\"{{ .MatchContext.Method }}\"}}"
    
mutators:
  noop:
    enabled: true

access_rules:
  repositories:
    - "file://./rules.json"
EOF
```
```
cat <<EOF > rules.json
    [{
    "id": "rule-1",
    "upstream": {
        "url": "http://localhost:8080"
    },
    "match": {
        "url": "http://<[^\/]+>/<.*>",
        "methods": ["GET","POST"]
    },
    "authenticators": [
        {
        "handler": "anonymous"
        },
        {
        "handler": "noop"
        }
    ],
    "authorizer": {
        "handler": "remote_opa"
    },
    "mutators": [
        {
        "handler": "noop"
        }
    ]
    }]
EOF
```

The intention of this authorizer is to be a drop in replacement to the python example in the OPA HTTP APIs documentation (https://www.openpolicyagent.org/docs/latest/http-api-authorization/). Rather than allow the policy input sent to the OPA REST API to be templated it is opinionated to use the format from the OPA documentation. An example policy input generated by the module looks like this:

```
{
  "input": {
    "user": "Subject",
    "path": ["path1", "path2", "path3"],
    "method": "GET"
  }
}
```
An example response from the OPA REST API looks like this:
```
{
  "result": {
    "allow": true
  }
}
```

## Checklist

- [ X ] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md).
- [ X ] I have read the [security policy](../security/policy).
- [ X ] I confirm that this pull request does not address a security
      the changes.
- [ X ] I have added tests that prove my fix is effective or that my feature
      works.
- [ X ] I have added or changed [the documentation](docs/docs)

## Further comments

This is the first Go project I've contributed to, more than happy for address feedback anyone has on any aspect of the code.


docs/docs/pipeline/authz.md
```diff
@@ -439,3 +439,98 @@ authorizers:
...
+
+- `remote` (string, required) - The remote OPA policy document URL located
+  under the /v1/data API path.
+
+#### Example
+
+```yaml
+# Global configuration file oathkeeper.yml
+authorizers:
+  remote_json:
```
aeneasr, line 470:
```suggestion
  remote_opa:
```

docs/docs/pipeline/authz.md
```diff
@@ -439,3 +439,98 @@ authorizers:
...
+      remote: http://opa-host:8181/v1/data/example/authz
+```
+
+```yaml
+# Some Access Rule: access-rule-1.yaml
+id: access-rule-1
+# match: ...
+# upstream: ...
+authorizers:
+  - handler: remote_json
```
aeneasr, line 484:
```suggestion
  - handler: remote_opa
```

docs/docs/pipeline/authz.md
```diff
@@ -439,3 +439,98 @@ authorizers:
...
+    "url": "http://my-app/api/<.*>",
+    "methods": ["GET"]
+  },
+  "authenticators": [
+    {
+      "handler": "anonymous"
+    }
+  ],
+  "authorizer": {
+    "handler": "remote_json",
```
aeneasr, line 506:
```suggestion
    "handler": "remote_opa",
```

pipeline/authz/remote_opa.go
```diff
@@ -0,0 +1,136 @@
...
+type OPAResponse struct {
+	Result struct {
+		Allow bool
+	}
+	Decision_id string
+}
+
+// OPA Policy Input
+type OPAInput struct {
+	User   string   `json:"user"`
```
aeneasr, line 53: This payload is bound to your model of access control and is not generalizable for other use cases. Instead, this should be configurable (see e.g. `remote_json` for an implementation). You might also want to wait for JsonNet support (https://github.com/ory/oathkeeper/issues/441) for this.

pipeline/authz/remote_opa.go
```diff
@@ -0,0 +1,136 @@
...
+	Input OPAInput `json:"input"`
+}
+
+// GetID implements the Authorizer interface.
+func (a *AuthorizerRemoteOPA) GetID() string {
+	return "remote_opa"
+}
+
+// Authorize implements the Authorizer interface.
+func (a *AuthorizerRemoteOPA) Authorize(_ *http.Request, session *authn.AuthenticationSession, config json.RawMessage, _ pipeline.Rule) error {
```
aeneasr, line 67: Needs a test :)

.schemas/authorizers.remote_opa.schema.json
```diff
@@ -0,0 +1,22 @@
+{
```
aeneasr, line 1: Revert, see https://github.com/ory/oathkeeper/blob/master/.schemas/README.md :) Changes should be made here: https://github.com/ory/oathkeeper/tree/master/.schema/

.schema/pipeline/authorizers.remote_opa.schema.json
```diff
@@ -0,0 +1,6 @@
+{
```
aeneasr, line 1: Revert, see https://github.com/ory/oathkeeper/blob/master/.schemas/README.md :) Changes should be made here: https://github.com/ory/oathkeeper/tree/master/.schema/

.schemas/config.schema.json
```diff
@@ -1468,6 +1468,38 @@
               }
             }
           ]
+        },
```
aeneasr, line 1471: Revert, see https://github.com/ory/oathkeeper/blob/master/.schemas/README.md :) Changes should be made here: https://github.com/ory/oathkeeper/tree/master/.schema/

pipeline/authz/remote_opa.go
ryanph: The payload for the OPA server needs the Method of the request, as well as the path of the request as an array. I believe I've added the Method correctly into the MatchContext structure, however I'm having a stab at the best way to add the Path "array" into the template.

Given it's specific to this module I don't think it makes sense to put it back in the AuthenticationSession struct but happy to take your guidance on that.

pipeline/authz/remote_opa.go
ryanph: I've basically copied the remote_json test, still reading up on how Viper works and how to run the tests locally to implement some more specific tests for this authorizer.

pipeline/authz/remote_opa.go
aeneasr: > The payload for the OPA server needs the Method of the request, as well as the path of the request as an array. I believe I've added the Method correctly into the MatchContext structure, however I'm having a stab at the best way to add the Path "array" into the template.

I don't think that is correct. You define the input model using Rego, so it depends on your Rego Policies what input you expect. You're probably referring to [this example](https://www.openpolicyagent.org/docs/latest/http-api-authorization/#2-load-a-policy-into-opa) but as you can see from the Rego policy, the input parameters are defined by the policy itself, not by OPA.

ryanph: Absolutely, the intention though was for it to be able to work with the OPA tutorial you linked "out of the box" and I've also seen this "path split into an array" in use by other developers I'm working with.

Any suggestions on how this could be achieved any other way using text/template, or do you have any objections to this approach?

aeneasr: I think the best approach would be using JsonNet (see https://github.com/ory/oathkeeper/issues/441) but I feat that needs quite some work on my end beforehand.

ryanph: Looks like there's no way to implement this in an acceptable way without JsonNet so I'll close off this PR and look at it again at some point in the future

Issue #467, opened 2020-06-15
Authenticate errors don't log the reason for the error


pike1212:
**Is your feature request related to a problem? Please describe.**

When JWT validation fails, there is nothing in the logs that say why (i.e. token expired)

**Describe the solution you'd like**

Add the error reason to that log statement

grantzvolsky:
To elaborate with a concrete example:

When Verifier returns from

https://github.com/ory/oathkeeper/blob/852fbc190f79c2848b4957dd945b1f4238fefbbb/credentials/verifier_default.go#L45

it correctly includes the reason. The reason is then lost in the Authenticator:

https://github.com/ory/oathkeeper/blob/852fbc190f79c2848b4957dd945b1f4238fefbbb/pipeline/authn/authenticator_jwt.go#L98

aeneasr:
Ah yeah, we should definitely pass that error there. If you have an idea how to do that feel free to PR :)

PR #468, opened 2020-06-15
feat: log error reason when auth fails

pike1212:
## Related issue

#467 

## Proposed changes

Add `error_reason` when authenticate fails

## Checklist

      the changes.
      works.


proxy/request_handler.go
```diff
@@ -229,6 +229,7 @@ func (d *RequestHandler) HandleRequest(r *http.Request, rl *rule.Rule) (session
 					WithField("granted", false).
 					WithField("authentication_handler", a.Handler).
 					WithField("reason_id", "authentication_handler_error").
+					WithField("error_reason", err.(*herodot.DefaultError).ReasonField).
```
aeneasr, line 232: I'm pretty sure this will panic at some point.

My plan was to upgrade ory/x here because we've much improved the logger in the new versions. It should include the reason field when logging the error now without any additional work.

PR #424, opened 2020-04-29
feat: cache incoming oauth introspect tokens

pike1212:
## Related issue

#293 

## Proposed changes

Adds ability to cache incoming oauth tokens in the introspector. By default will use the expires time from the token introspect endpoint to determine when to evict, or can override with a set TTL.

## Checklist

- [x] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md).
- [x] I have read the [security policy](../security/policy).
- [x] I confirm that this pull request does not address a security
      vulnerability. If this pull request addresses a security. vulnerability, I
      confirm that I got green light (please contact
      [security@ory.sh](mailto:security@ory.sh)) from the maintainers to push
      the changes.
- [ ] I have added tests that prove my fix is effective or that my feature
      works.
- [ ] I have added or changed [the documentation](docs/docs).

Issue #293, opened 2019-11-12
Caching in the oauth2_introspection authorizer


pike1212:
**Is your feature request related to a problem? Please describe.**

oauth2_introspection authorizer may have to make multiple requests for every incoming token (one for the introspection, the other to get a token if using the pre_authorization step)

**Describe the solution you'd like**

I would like Oathkeeper to cache both incoming tokens and the pre_authorization token

aeneasr:
Yeah that makes sense. Would you be up for a PR?

pike1212:
Yes, we are working on one.

aeneasr:
Awesome!

PR #482, opened 2020-07-15
feat: Cache OAuth2 introspection client

ptescher:
## Related issue

#293 
#424 

## Proposed changes

#424 Cached the responses of OAuth2 introspection, but for repeated calls to similar requests with pre_authorization enabled Oathkeeper will fetch a new access token per request. This can generate a massive number of access tokens!

## Checklist

      the changes.
      works.

## Further comments

Although the configuration for `authenticators. oauth2_introspection.config. pre_authorization` is not likely to change, the way the authenticators are built the configuration is not available on initialization. In order to get around this limitation this patch introduces a "cache" for the client. That cache uses a mutex which could be quite heavy for this use case. If there is a lighter solution that would work in this case it might be preferred.

### DIFF
--- a/pipeline/authn/authenticator_oauth2_introspection.go
+++ b/pipeline/authn/authenticator_oauth2_introspection.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/dgraph-io/ristretto"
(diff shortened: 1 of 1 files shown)


pipeline/authn/authenticator_oauth2_introspection.go
```diff
@@ -231,12 +260,7 @@ func (a *AuthenticatorOAuth2Introspection) Config(config json.RawMessage) (*Auth
 	var rt http.RoundTripper
 
 	if c.PreAuth != nil && c.PreAuth.Enabled {
-		rt = (&clientcredentials.Config{
-			ClientID:     c.PreAuth.ClientID,
-			ClientSecret: c.PreAuth.ClientSecret,
-			Scopes:       c.PreAuth.Scope,
-			TokenURL:     c.PreAuth.TokenURL,
-		}).Client(context.Background()).Transport
+		rt = a.preAuthCache.Client(context.Background(), c.PreAuth).Transport
```
aeneasr, line 263: `c.PreAuth` is a parameter which may change per route, but the authenticator is instantiated only once with one global cache, which will cause errors when more than one PreAuth are configured. Instead, this should probably be cached using the c.PreAuth contents as the key (e.g. crc32ing a string representation of c.PreAuth) .

Another issue is that the cache is never invalidated, even if requests start failing. This is problematic when, for example, the configuration is changed and subsequently reloaded.

tarjanik:
@aeneasr Is this feature resolved in any other way? For me it seems the pre_authorization is not usable without this in production. Thanks

PR #487, opened 2020-07-29
refactor: refactor decisions API and add traefik (#486)

aeneasr:
This is an umbrella PR for #441


PeteMac88:
Any new here?

Apologiz:
Hi, @aeneasr, I hope you are well!
Please tell me when do you plan to send this draft for release?

We are faced with the problem of matching Host from the incoming requests, which is why we would like to use a new endpoint for traefik, to which custom headers can be passed.

Thank!

aeneasr:
We're currently hiring developers to help with the refactoring as I don't have time to work on this - depending on that process we'll see how it goes. We probably want Oathkeeper as part of our cloud offering in 2022 so will be working on that latest then

designermonkey:
I'm really impressed with the software so far from Ory, and I wish I had golang code under my belt so I could offer some help here, sadly as I don't I'd like to simply offer some feedback instead, and I hope that's ok.

I think that attaching this to the specific term Traefik is a bad move that will come back to bite as more and more auth services may expect specific implementations.A more generic approach would better suit the goal and I feel that config may be better suited to deal with the different patterns that the decisions api can provide.

```yaml
---
serve:
  api:
    port: 4456
    decision_strategy: path|headers
    requestHeaders:
      host: 'X-Forwarded-Host'
      method: 'X-Forwarded-Method'
      protocol: 'X-Forwarded-Proro'
      uri: 'X-Forwarded-Uri'
```

How about something along these lines? This way if a custom auth service used different headers (like one I built a while back) it could be easily plugged in to Oathkeeper.

aeneasr:
I currently do not have time to work on this, closing it for now.

PR #489, opened 2020-07-29
feat: add trace to forwarded requests headers

err0r500:
## Proposed changes

Inject a new tracing span in headers if a request is forwarded to an upstream in order to link spans between services

NB : it relies on a method of ory/x/tracing.Tracer that has not been merged (but a PR has been submitted : https://github.com/ory/x/pull/179 )

## Checklist

      the changes.
      works.


aeneasr:
Thank you for this change! I am unfortunately not too familiar with tracing, could you explain the reasoning a bit? I'd also be interested to learn what the "standard" way of doing this is, preferably with some documentation to back it up. Thank you!

err0r500:
It's just adding some headers to a request so the receiver can unpack them, create a child span :

https://github.com/opentracing/opentracing-go#serializing-to-the-wire

For example, when it's activated on Kratos to, you can get the complete flow between oathkeeper and kratos : 
Screenshot from 2020-07-30 15-14-11

proxy/proxy.go
```diff
@@ -88,6 +91,22 @@ func (d *Proxy) RoundTrip(r *http.Request) (*http.Response, error) {
 			Header:     rw.header,
 		}, nil
 	} else if err == nil {
+		if d.t.IsLoaded() {
+			var sp opentracing.Span
+			spanName := "forwarding_to:" + r.URL.String()
```
aeneasr, line 96: Given that this would be many, many operations depending on the number of URLs, could it be a problem with cardinality? I was under the impression that one would use [semantic conventions for tags](https://opentracing.io/specification/conventions/) such as `http.url` and a span name which mimics the function name (in this case `round_trip` or something similar)

err0r500: you're right about semantic convention : if there's a standard, it's better to follow it.
about how to implement that, traefik has tracing capabilities so I'll have a look about how they did that and come back to you.

aeneasr: Thank you, that would be really helpful!

aeneasr:
Closing due to inactivity

Issue #488, opened 2020-07-29
Improve the docker-compose experience


err0r500:
It would be nice to be able to run `docker-compose up` from the root of the repo and get a working example.

I guess I'm not the only one to have experienced permission issues for config files so if it looks interesting to you, I've a PR ready to be submitted : https://github.com/ory/oathkeeper/compare/master...err0r500:docker-compose-improvement

Just tell me

err0r500:
actually, it's more a SELinux issue, I'm not sure about know how bind mounts with the `:Z` option behaves on other systems

PR #490, opened 2020-07-29
feat : Docker compose improvement

err0r500:
## Related issue
#488 

## Proposed changes

This PR should allow the docker-compose to run out of the box.

It uses a temporary container (busybox) to copy the content of `./.docker_compose/` into a named volume used by oathkeeper

## Checklist

      the changes.
      works.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,22 +1,13 @@
 version: '3.7'
 
 services:
-
-  volume-populate: 
-  # this is just used to copy the content of ./.docker_compose to oathkeeper-volume
-    image: busybox
-    command: cp -r /src/. /dest
-    volumes:
-      - ./.docker_compose:/src:Z
-      - oathkeeper-volume:/dest
-
   oathkeeper:
     build:
       context: .
       dockerfile: Dockerfile-dc
     ports:
-      - 4455:4455
-      - 4456:4456
+      - "4455:4455"
+      - "4456:4456"
     command:
       serve --config=/etc/config/oathkeeper/config.yaml
     environment:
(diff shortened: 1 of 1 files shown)


aeneasr:
What's the difference between using an intermediate volume to copy the files, and mounting the files directly? What type of permission issue do you encounter? Does it have to do with the user ID?

err0r500:
Hi,
As I mentioned in the related issue #490 it's a problem with SELinux and adding the `Z` flag is in fact enough.

I think the work around was working because SELinux doesn't care if you just copy files without trying to read/write them and once they were in the named volume, SELinux doesn't care at all of what you do with these files.

aeneasr:
Ok great, in that case it would probably make sense to document this limitation of SELinux somewhere around [here](https://github.com/ory/oathkeeper/blob/master/docs/docs/configure-deploy.md#configure), for example:

```
:::note

Make sure to have the `Z` flag enabled when running this with SELinux enabled.

:::
```

It would probably make sense to say how to enable the `Z` flag as well (chmod?) :)

err0r500:
Unfortunately, the `Z` flag must be set in the docker-compose file, like so : 

```
    volumes:
      - ./.docker_compose:/etc/config/oathkeeper:Z
```

here's the explanation : https://www.projectatomic.io/blog/2015/06/using-volumes-with-docker-can-cause-problems-with-selinux/

aeneasr:
Oh I see, so maybe explain that and link to the article?

err0r500:
It can be an option, yes. 

I was about to say it was not super user-friendly, especially if the change doesn't affect the other users... but since it looks like I'm the first one to hit the problem (and it's of course in all Ory's projects) SELinux must be disabled on pretty much everyone's machine :)

aeneasr:
Yeah I don't think this was reported anywhere else although we did have some issues with ORY Kratos but I think it was related to how the image was set up. If we can guarantee that the flag does not have any negative behavior on other platforms (macOS, not SELinux, windows) we can also just merge it as is. Do you have access to any of those OS?

err0r500:
yeap, I can find both, I'll tell you before the end of the day (Paris time)

aeneasr:
Closing due to inactivity

PR #514, opened 2020-09-14
fix windows path error

alexanyang:
this issue is happened on windows 10 system,it may cased by incorrect string parse with url.Parse() method .that split string with ' : ',and make it wrong . The error message is as follows：

time="2020-01-08T15:43:39+08:00" level=error msg="Unable to update access rules from given location, changes will be ignored. Check the configuration or restart the service if the issue persists." error="parse file://\F:\aweso
meProject\bin\resource-server.json: invalid port ":\\awesomeProject\\bin\\resource-server.json" after host" file="file:///F:/awesomeProject/bin/resource-server.json"

## Checklist

      the changes.
      works.

### DIFF
--- a/rule/fetcher_default.go
+++ b/rule/fetcher_default.go
@@ -144,14 +144,15 @@ func (f *FetcherDefault) configUpdate(ctx context.Context, watcher *fsnotify.Wat
 }
 
 func (f *FetcherDefault) sourceUpdate(e event) ([]Rule, error) {
-	if e.path.Scheme == "file" {
-		u, err := url.Parse("file://" + filepath.Clean(strings.TrimPrefix(e.path.String(), "file://")))
-		if err != nil {
-			return nil, err
-		}
-
-		e.path = *u
-	}
+// unnecessary code will make an error on windows machine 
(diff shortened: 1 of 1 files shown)


alexanyang:
if you need this fix ,just test and merge it

aeneasr (commented):
Thank you, given that this changeset removes a check and does not introduce new tests demonstrating the effectiveness and potential regressions, could you please add them so that we know this does what it is supposed to do? Thank you!

alexanyang:
> Thank you, given that this changeset removes a check and does not introduce new tests demonstrating the effectiveness and potential regressions, could you please add them so that we know this does what it is supposed to do? Thank you!

Sorry ,I have no time to give you a test example. That error caused by parse twice ,you can test with any URL likes `H:\Temp\authkeeper\src\github.com\ory`.
And if  `e.path.Scheme == "file"`  == true,that means you had parse once . Second url.Parse(...) will make error that will change value of that struct.

aeneasr:
Ok, I will have to check this on a windows machine. But without tests, this can't unfortunately be merged :/

Issue #505, opened 2020-08-26
Clean up logging in case of invalid credentials


catper:
**Is your feature request related to a problem? Please describe.**

When a request gets denied due to invalid credentials (and perhaps in other cases too but this is the only one that's bothering me atm) a lot of logs are produced and the information within them tends to overlap. I also think that the log level for some of them isn't really motivated. When something like this happens during load/soak tests, output really goes through the roof.

**Describe the solution you'd like**

It would be nice if the logs could be cleaned up a bit and the levels at which logs are emitted were re-evaluated.

**Additional context**
Example output:

```
2020-08-26T13:23:35.836628111Z time="2020-08-26T13:23:35Z" level=warning msg="The authentication handler encountered an error" authentication_handler=oauth2_introspection error="Access credentials are invalid" granted=false http_host=test.istio-ingress.com http_method=GET http_url="http://test.istio-ingress.com/v1/greetings" http_user_agent= reason_id=authentication_handler_error rule_id=authenticate
2020-08-26T13:23:35.836636521Z time="2020-08-26T13:23:35Z" level=warning msg="Access request denied" error="Access credentials are invalid" granted=false http_host=test.istio-ingress.com http_method=GET http_url="http://test.istio-ingress.com/v1/greetings" http_user_agent=
2020-08-26T13:23:35.836674487Z time="2020-08-26T13:23:35Z" level=error msg="An error occurred while handling a request" code=401 debug= details="map[]" error="The request could not be authorized" reason= request-id=b62bdf80-a6bc-4c9a-a8dc-9c47ccde8fd4 status=401 writer=JSON
2020-08-26T13:23:35.836709377Z time="2020-08-26T13:23:35Z" level=info msg="completed handling request" measure#oathkeeper-api.latency=5958530 method=GET remote="127.0.0.1:41592" request=/decisions/v1/greetings request_id=b62bdf80-a6bc-4c9a-a8dc-9c47ccde8fd4 status=401 text_status=Unauthorized took=5.95853ms
````

The first two messages, in my opinion, basically say the same thing and could be condensed into one. I also think they belong on `info` level, not `warning` as they don't actually indicate a problem with the application, just with the request.

[...]

PR #517, opened 2020-09-15
feat: Log invalid credentials on info level instead of error/warning

catper:
## Related issue
[#505 ](https://github.com/ory/oathkeeper/issues/505)

## Proposed changes

Invalid credentials or a malformed request is really the client's responsibility and should therefore not be considered an error.

## Checklist

      the changes.
      works.

## Further comments

In the best of worlds I would also like to not have an error logged in the scenario where e.g. the token was expired:

``` 
level=error msg=An error occurred while handling a request code=401 debug= details=map[] error=The request could not be authorized reason= request-id=194689ec322a9ad1a71ae5 ...
```

but that seems to be rather a large change. Is there any easy way to accomplish this?

### DIFF
--- a/proxy/request_handler.go
+++ b/proxy/request_handler.go
@@ -225,6 +225,7 @@ func (d *RequestHandler) HandleRequest(r *http.Request, rl *rule.Rule) (session 
 			// return nil
 			case helper.ErrUnauthorized.ErrorField:
 				d.r.Logger().Info(err)
+				return nil, err
 			default:
 				d.r.Logger().WithError(err).
 					WithFields(fields).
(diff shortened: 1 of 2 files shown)


proxy/request_handler.go
```diff
@@ -242,11 +244,15 @@ func (d *RequestHandler) HandleRequest(r *http.Request, rl *rule.Rule) (session
...
 	if !found {
 		err := errors.WithStack(helper.ErrUnauthorized)
-		d.r.Logger().WithError(err).
-			WithFields(fields).
-			WithField("granted", false).
-			WithField("reason_id", "authentication_handler_no_match").
-			Warn("No authentication handler was responsible for handling the authentication request")
+		// found will only be true if an authenticator was found AND authorisation was granted so check
+		// that this wasn't just a problem with the request
+		if err.Error() != helper.ErrUnauthorized.ErrorField {
```
aeneasr, line 249: This is always false because we set the error to `err := errors.WithStack(helper.ErrUnauthorized)` right before - so this can be removed :)

aeneasr (changes requested):

proxy/request_handler.go
catper: i see your point :), guess i need to think about this a bit more.

the problem is, if this if statement is removed, then invalid credentials or an expired token will result in `level=warning msg=No authentication handler was responsible for handling the authentication request` getting logged, which is inaccurate and misleading.

aeneasr (approved):
Awesome 🎉 

Thank you for your contribution!

PR #530, opened 2020-09-29
fix: check content-length header in lowercase

flusflas:
Issue #422 didn't fix the problem with the requests' `Content-Length` being copied in the responses because the check was case-sensitive and unit tests didn't cover it (now it does).


aeneasr (approved):
Awesome 🎉 

Thank you for your contribution!

Issue #528, opened 2020-09-29
Enable forwarding of original authorization header to (remote) authorizer


catper:
**Is your feature request related to a problem? Please describe.**

Authorization is typically done on the subject but there may be additional information in the access token that is needed for a decision to be made. At present, it doesn't seem possible to forward the original access token to a remote authorizer though, you can only forward data that exists within the authSession object can be passed on.

**Describe the solution you'd like**

I'd like the auth header to be forwarded to the remote authorizer.

aeneasr:
Thank you! This makes definite sense and will be part of https://github.com/ory/oathkeeper/issues/441 once we work on it :)

catper:
While waiting for #441 to happen, would it be acceptable to implement this as "stand alone" feature? I'm thinking all that's needed is a single line, copying the auth header and then we'd be done?

ashneverdawn:
I'm interested in this too.

aeneasr:
Yes, sounds good! Open to contributions! :)

PR #554, opened 2020-10-08
feat: Forward original authorization header when using remote (json) authorizer

catper:
## Related issue

#528 

## Proposed changes

Sometimes the user/subject from the authSession is not enough when attempting to make informed access control decisions. This PR grants access to the authorization header from the original request when using the `remote` and `remote_json` authorizers so that additional information can be extracted from the token if need be.

## Checklist

      the changes.
      works.

## Further comments

n/a

### DIFF
--- a/pipeline/authz/remote_json.go
+++ b/pipeline/authz/remote_json.go
@@ -84,7 +84,10 @@ func (a *AuthorizerRemoteJSON) Authorize(r *http.Request, session *authn.Authent
 		return errors.WithStack(err)
 	}
 	req.Header.Add("Content-Type", "application/json")
-	req.Header.Add("Authorization", r.Header.Get("Authorization"))
+	authz := r.Header.Get("Authorization")
+	if authz != "" {
+		req.Header.Add("Authorization", authz)
+	}
 
 	res, err := a.client.Do(req)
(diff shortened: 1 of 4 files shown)


pipeline/authz/remote_json.go
```diff
@@ -84,6 +84,7 @@ func (a *AuthorizerRemoteJSON) Authorize(_ *http.Request, session *authn.Authent
 		return errors.WithStack(err)
 	}
 	req.Header.Add("Content-Type", "application/json")
+	req.Header.Add("Authorization", r.Header.Get("Authorization"))
```
zepatrik, line 87: What happens when there is no `Authorization` header in the original request? Looks like it will still be set, then to an empty string?

catper: mmm, that's not so nice i guess. changed to only forward if authz header is non-empty

zepatrik: can you maybe also add a test case for that? thanks :wink:

catper: done!

zepatrik (approved):
Nice, looking good :+1:

aeneasr (approved):

Issue #553, opened 2020-10-08
Authenticator oauth2_introspection only works if token_type returned is an "access_token"


daviddelucca:
**Describe the bug**

When the introspect URL returns a token_type that is not an `access_token`, Oathkeeper throws an exception:

```
Introspected token is not an access token but "..."
```

This error is related to this [validation](https://github.com/ory/oathkeeper/blob/master/pipeline/authn/authenticator_oauth2_introspection.go#L171-L173).

**Reproducing the bug**

_Server logs_

```
time="2020-10-07T20:41:47Z" level=error msg="An error occurred while handling a request" code=403 debug= details="map[]" error="Access credentials are not sufficient to access this resource" reason="Introspected token is not an access token but \"Bearer\"" request-id=53e06937-f10a-4164-9f6e-d1a2ef2224fd status=403 writer=JSON
```

**Expected behavior**

It should accept other token_types instead of works only with access_token.

**Environment**

- Version: v0.38.2-beta.1
- Environment: Ory Helm Chart 

**Additional context**

If the solution is simple like remove that [validation](https://github.com/ory/oathkeeper/blob/master/pipeline/authn/authenticator_oauth2_introspection.go#L171-L173) or validate the `token_type` accross a few values, I could open a PR for it. I have read [RFC](https://tools.ietf.org/html/rfc7662#page-4) and I didn't find what are the expected values or if there are any.

aeneasr:
Yes we should change that - the token_type is Bearer. We'll be introducing a new `token_use` in the introspection response which would then be access token for example. I think we should fix this here!

aeneasr:
Would you be up to contributing that change?

daviddelucca:
I would like to contribute but I didn't understand what do you mean by 

> Yes we should change that - the token_type is Bearer. We'll be introducing a new token_use in the introspection response which would then be access token for example. I think we should fix this here!

Will you include the token_use in a newer version and I could just remove that validation?

PR #556, opened 2020-10-11
fix: remove token_type validation from authenticator_oauth2 introspec…

daviddelucca:
The authenticator_oauth2 introspection flow only works if the `token_type` is an `access_token`. It works great if your authentication server always return this type but it crashes if returns other types.

## Related issue

https://github.com/ory/oathkeeper/issues/553

## Proposed changes

My proposal is to remove this validation and allow other `token_type`'s. I have checked RFC but I didn't find what types should be allowed so my suggestion is to allow any token_type.

## Checklist

      the changes.
- [X] I have added tests that prove my fix is effective or that my feature
      works.

## Further comments

There is no docs to update.

### DIFF
--- a/pipeline/authn/authenticator_oauth2_introspection.go
+++ b/pipeline/authn/authenticator_oauth2_introspection.go
@@ -90,7 +90,7 @@ type AuthenticatorOAuth2IntrospectionResult struct {
 	ClientID  string                 `json:"client_id,omitempty"`
 	Scope     string                 `json:"scope,omitempty"`
 	Expires   int64                  `json:"exp"`
-	TokenUse string                 `json:"token_use"`
+	TokenUse  string                 `json:"token_use"`
 }
 
 func (a *AuthenticatorOAuth2Introspection) tokenFromCache(config *AuthenticatorOAuth2IntrospectionConfiguration, token string) (*AuthenticatorOAuth2IntrospectionResult, bool) {
--- a/pipeline/authn/authenticator_oauth2_introspection.go
+++ b/pipeline/authn/authenticator_oauth2_introspection.go
@@ -90,6 +90,7 @@ type AuthenticatorOAuth2IntrospectionResult struct {
 	ClientID  string                 `json:"client_id,omitempty"`
 	Scope     string                 `json:"scope,omitempty"`
 	Expires   int64                  `json:"exp"`
+	TokenUse string                 `json:"token_use"`
 }
 
 func (a *AuthenticatorOAuth2Introspection) tokenFromCache(config *AuthenticatorOAuth2IntrospectionConfiguration, token string) (*AuthenticatorOAuth2IntrospectionResult, bool) {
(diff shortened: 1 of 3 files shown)


pipeline/authn/authenticator_oauth2_introspection.go
```diff
@@ -168,10 +168,6 @@ func (a *AuthenticatorOAuth2Introspection) Authenticate(r *http.Request, session
 			return errors.WithStack(err)
 		}
 
-		if len(i.TokenType) > 0 && i.TokenType != "access_token" {
-			return errors.WithStack(helper.ErrForbidden.WithReason(fmt.Sprintf("Introspected token is not an access token but \"%s\"", i.TokenType)))
-		}
```
aeneasr, line 173: So ORY Hydra 1.9.0 introduces a new parameter `token_use` which can be `access_token` or `refresh_token`. I think we should keep this method here but add `TokenUse string                 `json:"token_use"`` [here](https://github.com/ory/oathkeeper/pull/556/files#diff-5e85f382e854a595d5161855726166f6R88).

It would be nice if you could also add a test for this!

aeneasr (changes requested):
Awesome, thank you for your contribution! This looks pretty good but I have some ideas how to improve it further :)

pipeline/authn/authenticator_oauth2_introspection.go
daviddelucca: Got it! We will check for `token_use` instead of `token_type`. I'll update this PR.

daviddelucca: Could you check if is [it](https://github.com/ory/oathkeeper/pull/556/files#diff-5c8061290da17d352a4b888e5270cdd071ea584cd4751011132d8cccd69594f0R172-R174) what you were talking about?

pipeline/authn/authenticator_oauth2_introspection.go
```diff
@@ -168,8 +169,8 @@ func (a *AuthenticatorOAuth2Introspection) Authenticate(r *http.Request, session
 			return errors.WithStack(err)
 		}
 
-		if len(i.TokenType) > 0 && i.TokenType != "access_token" {
-			return errors.WithStack(helper.ErrForbidden.WithReason(fmt.Sprintf("Introspected token is not an access token but \"%s\"", i.TokenType)))
+		if len(i.TokenUse) > 0 && (i.TokenUse != "access_token" && i.TokenUse != "refresh_token") {
```
aeneasr, line 172: Close :) We only want to allow `access_tokens` here as they are for access of APIs, and refresh_tokens are for refreshing access tokens :)

aeneasr (changes requested):

pipeline/authn/authenticator_oauth2_introspection.go
daviddelucca: :D
Got it and fixed 👯

aeneasr (approved):
Thank you for all the hard work!

PR #557, opened 2020-10-15
fix: Read windows path error - 2

PatrLind:
## Related PR: https://github.com/ory/oathkeeper/pull/514
I asked in the Slack chat if I should make this PR and it seems like @zepatrik and @jfcurran gave their permission with an emoji.

## Proposed changes
I had the same issue as the person creating PR 514. That file:// URLs wasn't working correctly under Windows.
I looked into it a bit and it seems like the original author of the PR and also Ory have misunderstood how file:// URLs work.
I have added a helper function that can extract the file path from the URL for as many situations I can think of.
I have also added a few non standard ways (relative paths and incomplete URLs) to deal with file:// URLs because those were used in the original tests from Ory.

## Checklist
      the changes.
- [X] I have added tests that prove my fix is effective or that my feature
      works.

## Further comments
The tests didn't work on Windows before my fix, but now they work except for a test called TestFetcherWatchRepositoryFromKubernetesConfigMap. I am not sure how to make this test succeed since it seems to be written for non windows systems only. I have also run the tests under WSL (Linux) and they all pass, so I think this change should not cause any regressions.

I don't particularly like all the workarounds in the helper function, but it seems like it is common for people to write incorrect file:// URLs so I think they are needed.


zepatrik:
The kubernetes test does not have to pass as you would not run kubernetes on windows. They do some symlink magic there to update files atomically, that's why we have that special case.

helper/fileurl_test.go
```diff
@@ -0,0 +1,38 @@
+package helper
```
zepatrik, line 1: we have the x package for project wide helpers :wink: 
```suggestion
package x
```

helper/fileurl.go
```diff
@@ -0,0 +1,48 @@
+package helper
```
zepatrik, line 1:
```suggestion
package x
```

zepatrik (changes requested):
It would actually be very nice if you could add a CIrcleCI job or GitHub workflow to run the tests on windows. I think it would be enough to run `go test ./...`. If you have no experience with that, let me know.

PatrLind:
I don't have any CIrcleCI or GitHub Workflow experience at all, but I can give it a try.

PatrLind:
Ok, So I was not successful in doing a Windows CircleCI test, not sure how to do it. Sorry.
Also, how come the existing CircleCI tests are failing?

zepatrik:
The error output is:
```
panic: proxy: expected 200: {"error":{"code":404,"status":"Not Found","message":"The requested resource could not be found"}}


goroutine 1 [running]:
main.main()
	/go/src/github.com/ory/oathkeeper/test/e2e/okclient/main.go:57 +0x22d
+ finish
+ cat ./oathkeeper.e2e.log
time=2020-10-15T09:52:52Z level=info msg=Config file loaded successfully. audience=application path=./config.yml service_name=oathkeeper service_version=
Thank you for using ORY Oathkeeper master!

Take security seriously and subscribe to the ORY Security Newsletter. Stay on top of new patches and security insights.                                                                                                

>> Subscribe now: http://eepurl.com/di390P <<
time=2020-10-15T09:52:52Z level=debug msg=Viper detected a configuration change, updating matching strategy audience=application event=matching_strategy_config_change service_name=ORY Oathkeeper service_version=master source=entrypoint
time=2020-10-15T09:52:52Z level=debug msg=Viper detected a configuration change, reloading config. audience=application event=config_change service_name=ORY Oathkeeper service_version=master source=entrypoint
time=2020-10-15T09:52:52Z level=debug msg=One or more access rule repositories changed, reloading access rules. audience=application event=repository_change file=file://e2e-rules.json service_name=ORY Oathkeeper service_version=master source=config_update
time=2020-10-15T09:52:52Z level=debug msg=Fetching access rules from given location because something changed. audience=application location=file://e2e-rules.json service_name=ORY Oathkeeper service_version=master
time=2020-10-15T09:52:52Z level=error msg=Unable to update access rules from given location, changes will be ignored. Check the configuration or restart the service if the issue persists. audience=application error=map[message:error converting YAML to JSON: yaml: did not find expected alphabetic or numeric character] file=file://e2e-rules.json service_name=ORY Oathkeeper service_version=master
time=2020-10-15T09:52:52Z level=info msg=Software quality assurance features are enabled. Learn more at: https://www.ory.sh/docs/ecosystem/sqa audience=application service_name=ORY Oathkeeper service_version=master
time=2020-10-15T09:52:52Z level=info msg=No tracer configured - skipping tracing setup audience=application service_name=ORY Oathkeeper service_version=master
time=2020-10-15T09:52:52Z level=info msg=Listening on http://:9000 audience=application service_name=ORY Oathkeeper service_version=master
time=2020-10-15T09:52:52Z level=info msg=TLS has not been configured for proxy, skipping audience=application service_name=ORY Oathkeeper service_version=master
time=2020-10-15T09:52:52Z level=info msg=TLS has not been configured for api, skipping audience=application service_name=ORY Oathkeeper service_version=master
time=2020-10-15T09:52:52Z level=info msg=Listening on http://:6661 audience=application service_name=ORY Oathkeeper service_version=master
time=2020-10-15T09:52:52Z level=info msg=Listening on http://:6660 audience=application service_name=ORY Oathkeeper service_version=master
time=2020-10-15T09:52:53Z level=info msg=started handling request http_request=map[headers:map[accept-encoding:gzip authorization:Value is sensitive and has been redacted. To see the value set config key "log.leak_sensitive_values = true" or environment variable "LOG_LEAK_SENSITIVE_VALUES=true". user-agent:Go-http-client/1.1] host:127.0.0.1:6660 method:GET path:/jwt query:<nil> remote:127.0.0.1:58560 scheme:http]
time=2020-10-15T09:52:53Z level=warning msg=Access request denied audience=application error=map[debug: message:Requested url does not match any rules reason: status:Not Found status_code:404] granted=false http_host=127.0.0.1:6660 http_method=GET http_url=http://127.0.0.1:6660/jwt http_user_agent=Go-http-client/1.1 service_name=ORY Oathkeeper service_version=master
time=2020-10-15T09:52:53Z level=error msg=An error occurred while handling a request code=404 debug= details=map[] error=The requested resource could not be found reason= request-id= status=404 writer=JSON
time=2020-10-15T09:52:53Z level=info msg=completed handling request http_request=map[headers:map[accept-encoding:gzip authorization:Value is sensitive and has been redacted. To see the value set config key "log.leak_sensitive_values = true" or environment variable "LOG_LEAK_SENSITIVE_VALUES=true". user-agent:Go-http-client/1.1] host:127.0.0.1:6660 method:GET path:/jwt query:<nil> remote:127.0.0.1:58560 scheme:http] http_response=map[status:404 text_status:Not Found took:2.852785ms]
+ echo -----
-----
+ cat ./api.e2e.log

```

PatrLind:
Ok, I should have been more clear, I can see the logs but I am not sure what part of the output is the error that fails the test.
There are three errors, none of which seems to be related to what I did:
* panic: proxy: expected 200: {"error":{"code":404,"status":"Not Found","message":"The requested resource could not be found"}}
* level=error msg=Unable to update access rules from given location, changes will be ignored. Check the configuration or restart the service if the issue persists. audience=application error=map[message:error converting YAML to JSON: yaml: did not find expected alphabetic or numeric character] file=file://e2e-rules.json service_name=ORY Oathkeeper service_version=master
* level=error msg=An error occurred while handling a request code=404 debug= details=map[] error=The requested resource could not be found reason= request-id= status=404 writer=JSON

All the relevant test seems to pass when I run them locally, but this log doesn't really run those tests at all. Maybe I need to run some other test script locally to find out where the issue is? As I am not so familiar with your build/test system some help here would be appreciated.

helper/fileurl_test.go
PatrLind: I have moved this file to the x package

helper/fileurl.go
PatrLind: I have moved this file to the x package

PatrLind:
Please note that I have made some small change to rule/rule_test.go that corrected an issue that made the tests (or was it linting?) fail. I am not sure this change is allowed or should be in here. Should I remove it?

zepatrik (changes requested):
Can you re-request a review when your done? Thx :wink:

PatrLind:
Ok, the tests (except the Windows test) are passing now. It was quite a bit more involved than I had hoped!
The reason is because the ffile URLs are involved in a lot of components and tests.

In the end I had to separate the fix in to two parts.
The first part handles the parsing of URLs in general and has special handling of parsing file:// URLs. The reason is because the standard library url.Parse() doesn't handle all the non standard cases. I noticed that you are using github.com/ory/x/urlx to handle some of this before, but I felt I couldn't just go and change the code in there, so I duplicated it into the local x package.
Not sure this is what is best, but at least it allows me to demonstrate it.
The output of the first part is a *url.URL that have a standardized format. Note however that the Scheme of the URL can be either file or an empty string. This is because the file scheme doesn't allow for relative paths. The Go url.Parse() function can deal with relative paths if the scheme is empty. Therefore anyone that uses these URL objects and are looking for file paths need to look for a scheme set to "file" OR "".

The second part is a function that can take a standard *url.URL object and return a path that is correct for the current system.
On POSIX systems this just returns the .Path component. On Windows the path is converted into either a normal path (C:\something) or a UNC network path (\\\\server\share\path).  The path delimiters are also converted.

If this change ends up being approved I think that this functionality probably should be moved over to github.com/ory/x/urlx to that it can be reused in other projects. 

Regarding the failing Windows test I don't think it is related to this PR, I get the same error when running the TestMutatorHydrator test from the github.com/ory/oathkeeper main branch.

What do you think?

zepatrik:
Nice writeup, thank you for your work :+1: :heart_eyes: 
We use the packages from `ory/x` in all of our projects, so if you fix it there it will actually work in other projects as well. Can you open a PR there? It will also be easier to run windows tests there :wink: We can remove them here again if it is tested in `x`.

x/fileurl.go
```diff
@@ -0,0 +1,110 @@
...
+		// Normally the first part after file:// is a hostname, but since
+		// this is often misused we interpret the URL like a normal path
+		// by removing the "file://" from the beginning
+		rawurl = rawurl[7:]
+	}
+	if winPathRegex.MatchString(rawurl) {
+		// Windows path
+		return url.Parse("file:///" + toSlash(rawurl))
+	}
+	if strings.Index(lcRawurl, "\\\\") == 0 {
```
zepatrik, line 58:
```suggestion
	if strings.HasPrefix(lcRawurl, "\\\\") {
```

x/fileurl.go
```diff
@@ -0,0 +1,110 @@
...
+		// Windows UNC path
+		// We extract the hostname and creates an appropriate file:// URL
+		// based on the hostname and the path
+		parts := strings.Split(filepath.FromSlash(rawurl), "\\")
```
zepatrik, line 62: On POSIX, `filepath.FromSlash` will result in a path with slashes again and the split will match literal backslashes (not that any sane person would use backslashes in directory/file names but it should not match those...)
I would say also the condition at L58 should probably not be true on POSIX. I feel like this might open an attack vector 10 years in the future where somebody uses `\` in filenames to circumvent some check.

x/fileurl.go
```diff
@@ -0,0 +1,110 @@
...
+		host := ""
+		if len(parts) > 2 {
+			host = parts[2]
+		}
+		p := "/"
+		if len(parts) > 4 {
+			p += strings.Join(parts[3:], "/")
+		}
```
zepatrik, line 70: This took me quite some time to understand this part. Can you maybe rewrite it a bit? Or add comments? It is especially not clear why `parts[3]` is discarded when `len(parts) == 4`.

x/fileurl.go
```diff
@@ -0,0 +1,110 @@
...
+		return url.Parse("file://" + host + p)
+	}
+	u, err := url.Parse(rawurl)
+	if err != nil {
+		return nil, err
+	}
+	if u.Scheme == "file" || u.Scheme == "" {
+		u.Path = toSlash(u.Path)
```
zepatrik, line 78: Here also, on POSIX `\` could be part of the file name and the conversion is not necessary at all.

x/fileurl.go
```diff
@@ -0,0 +1,110 @@
...
+func ParseOrFatal(l *logrusx.Logger, in string) *url.URL {
+	out, err := ParseURL(in)
+	if err != nil {
+		l.WithError(err).Fatalf("Unable to parse url: %s", in)
+	}
+	return out
+}
+
+func stripFistPathSeparators(fPath string) string {
+	for len(fPath) > 0 && (fPath[0] == '/' || fPath[0] == '\\') {
```
zepatrik, line 102: Here also, on POSIX `\` could be part of the file name.

x/fileurl_test.go
```diff
@@ -0,0 +1,102 @@
...
+		{"filE://../test/update/file4.txt", "../test/update/file4.txt", "../test/update/file4.txt"},
+		{"file://C:/users/test/file5.txt", "/C:/users/test/file5.txt", "file:///C:/users/test/file5.txt"},  // We expect a initial / in the path because this is a Windows absolute path
+		{"file:///C:/users/test/file6.txt", "/C:/users/test/file6.txt", "file:///C:/users/test/file6.txt"}, // --//--
+		{"file://file7.txt", "file7.txt", "file7.txt"},
+		{"file://path/file8.txt", "path/file8.txt", "path/file8.txt"},
+		{"file://C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\9ccf9f68-121c-451a-8a73-2aa360925b5a386398343/access-rules.json", "/C:/Users/RUNNER~1/AppData/Local/Temp/9ccf9f68-121c-451a-8a73-2aa360925b5a386398343/access-rules.json", "file:///C:/Users/RUNNER~1/AppData/Local/Temp/9ccf9f68-121c-451a-8a73-2aa360925b5a386398343/access-rules.json"},
+		{"file:///C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\9ccf9f68-121c-451a-8a73-2aa360925b5a386398343/access-rules.json", "/C:/Users/RUNNER~1/AppData/Local/Temp/9ccf9f68-121c-451a-8a73-2aa360925b5a386398343/access-rules.json", "file:///C:/Users/RUNNER~1/AppData/Local/Temp/9ccf9f68-121c-451a-8a73-2aa360925b5a386398343/access-rules.json"},
```
zepatrik, line 78: I am not quite sure this is what we want. Is this following some spec or standard?

x/fileurl.go
```diff
@@ -0,0 +1,110 @@
...
+			// is a normal path (for example: "\c:\" -> "c:\"")
+			fPath = stripFistPathSeparators(fPath)
+		}
+	}
+	return filepath.FromSlash(fPath)
+}
+
+// ParseURL parses rawurl into a URL structure with special handling for file:// URLs
+func ParseURL(rawurl string) (*url.URL, error) {
+	lcRawurl := strings.ToLower(rawurl)
```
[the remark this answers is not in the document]
zepatrik: Wouldn't it be less error prone to do
```suggestion
        if runtime.GOOS == "windows" {
	    rawURL = strings.ToLower(rawurl)
	}
```
and avoid the decision what variable to use for what check?

x/fileurl.go
```diff
@@ -0,0 +1,110 @@
...
+// winPathRegex is a regex for [DRIVE-LETTER]:
+var winPathRegex = regexp.MustCompile("^[A-Za-z]:.*")
+
+// GetURLFilePath returns the path of a URL that is compatible with the runtime os filesystem
+func GetURLFilePath(u *url.URL) string {
+	if u == nil {
+		return ""
+	}
+
+	if u.Scheme != "file" && u.Scheme != "" {
```
zepatrik, line 22: Let's invert this condition to make it more readable
```suggestion
	if u.Scheme == "file" || u.Scheme == "" {
```

zepatrik: Hi Mr. De Morgan :wave:

zepatrik (changes requested):

x/fileurl.go
```diff
@@ -0,0 +1,110 @@
...
+			// On Windows we should remove the initial path separator in case this
+			// is a normal path (for example: "\c:\" -> "c:\"")
+			fPath = stripFistPathSeparators(fPath)
+		}
+	}
+	return filepath.FromSlash(fPath)
+}
+
+// ParseURL parses rawurl into a URL structure with special handling for file:// URLs
+func ParseURL(rawurl string) (*url.URL, error) {
```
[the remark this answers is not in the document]
PatrLind: The reason for the name was because I copied it from the standard library, but I agree that the standard naming convention is better.

x/fileurl.go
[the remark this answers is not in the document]
PatrLind: runtime.GOOS is guaranteed to always be lower case, so I don't think this is needed. 
If you run `go tool dist list` you will see a list of all combinations of runtime.GOOS and runtime.GOARCH.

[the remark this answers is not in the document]
zepatrik: you are lower casing `rawURL` there because windows filepaths are case insensitive right? But they are case sensitive on POSIX
That was what I was referring to.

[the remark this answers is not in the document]
PatrLind: Not sure exactly what you mean here, the lcRawURL is only used to compare the initial part if the rawURL string so see if it starts with file://  (or any upper/lower case version of it).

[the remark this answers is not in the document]
zepatrik: Ah ok, could you then maybe use `strings.Split(rawURL, "://")` to get the scheme and use that for the checks?

x/fileurl.go
```diff
@@ -0,0 +1,110 @@
...
+	if strings.Index(lcRawurl, "file:///") == 0 {
+		return url.Parse("file:///" + toSlash(rawurl[8:]))
+	}
+	if strings.Index(lcRawurl, "file://") == 0 {
+		// Normally the first part after file:// is a hostname, but since
+		// this is often misused we interpret the URL like a normal path
+		// by removing the "file://" from the beginning
+		rawurl = rawurl[7:]
+	}
+	if winPathRegex.MatchString(rawurl) {
```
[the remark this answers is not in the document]
PatrLind: I think so because this is about parsing what is in the config file, and the config file can look the same on any system. It is only when we want to use the Path from the URL that the platform becomes a factor. However, when a POSIX system tries to use such a path it will probably return an error saying it cannot find /C:/path/file.ext. I think this is reasonable to expect.

x/fileurl.go
PatrLind: I will fix this

x/fileurl.go
PatrLind: Good call! I didn't think of this. I will make some tests and try to fix it

x/fileurl.go
PatrLind: I will fix

x/fileurl.go
zepatrik: Nice :+1:

x/fileurl_test.go
PatrLind: Well, this can be discussed of course, but I have determined that this is a reasonable solution since this style of non standard file URLs are used throughout Oathkeeper. And since the file scheme doesn't support relative paths the solution is to use an empty scheme. This is also documented in the standard library documentation: https://golang.org/pkg/net/url/#Parse
I have tried to combine the standard and also the way many people are probably using file URLs already without having any regressions.

x/fileurl.go
```diff
@@ -0,0 +1,110 @@
...
+	lcRawurl := strings.ToLower(rawurl)
+	if strings.Index(lcRawurl, "file:///") == 0 {
+		return url.Parse("file:///" + toSlash(rawurl[8:]))
+	}
+	if strings.Index(lcRawurl, "file://") == 0 {
+		// Normally the first part after file:// is a hostname, but since
+		// this is often misused we interpret the URL like a normal path
+		// by removing the "file://" from the beginning
+		rawurl = rawurl[7:]
+	}
```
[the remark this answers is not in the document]
PatrLind: No ://whatever will not parse. The error will be: `parse "://": missing protocol scheme`

x/fileurl.go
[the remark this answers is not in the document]
zepatrik: Yes, if this results in `/C:/path/file.ext` being used as the file path then that is what I would expect as well :+1:

x/fileurl.go
PatrLind: I have reworked this part now

x/fileurl.go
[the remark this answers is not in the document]
PatrLind: At this point in the code I am first looking for file:/// (three slashes), if that is the case I am parsing it as a normal URL with an early exit. If that is not the case and it instead starts with file:// (two slashes) we trim the first 7 chars off the beginning and then continues the function. The lower case comparison has nothing to do with Windows, it just makes sure that FILE:// and file:// are treated the same.
I can change it to a regexp case insensitive comparison if that feels better.

x/fileurl.go
PatrLind: I think I have fixed this issue now

x/fileurl.go
PatrLind: This is fixed

PatrLind:
So I have made some changes after your suggestions now.
As for the runtime.GOOS == "windows", I will address this in the urlx package by adding the appropriate build tags for the different versions.

x/fileurl.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// ParseURL parses rawURL into a URL structure with special handling for file:// URLs
+func ParseURL(rawURL string) (*url.URL, error) {
+	lcRawURL := strings.ToLower(rawURL)
+	if strings.HasPrefix(lcRawURL, "file:///") {
+		return url.Parse(rawURL)
+	}
+
+	if strings.HasPrefix(lcRawURL, "file://") {
```
zepatrik, line 49: I just realized you can avoid the `if` statement and use `strings.TrimPrefix(lcRawURL, "file://")`, sry :sweat_smile:

zepatrik (commented):
Nice, this looks good now :tada: 
Let's merge it in `ory/x`

x/fileurl.go
PatrLind: I has the same though, but it turns out that we are comparing the lower case converted string but modifying the non converted string so unfortunately that won't work. Unless there is a way to TrimPrefix without case sensitivity?

zepatrik: Ah no, you are actually trimming `rawURL`, nevermind.

zepatrik:
What is the current state of this @PatrLind ? Do you think it is enough to bump the `ory/x` version for oathkeeper or are any additional changes needed?

PatrLind:
I have been a bit busy on other things so this was kind of forgotten.
I have updated the ory/x version now. The only problem seems to be the failing Windows test:
TestFetcherWatchRepositoryFromFS/case=2
I am not sure why, is it related to this you think?

zepatrik:
Hm, in essence this means that file watching does not work on windows? Although it worked for the first case. Maybe it is flaky? Can you reproduce that locally? I don't have a windows machine at hand.

zepatrik:
Damn, the windows tests are flaky AF... I just reran it and `rules` passed now but other tests are failing now.

zepatrik:
Github actions overwrites logs on rerun :sleepy: here are at least some logs I could gather:

- [authn and mutate failing](https://github.com/ory/oathkeeper/files/5554304/3_Run.go.test.-failfast.-timeout.20m.txt)
- [mutate failing](https://github.com/ory/oathkeeper/files/5554291/3_Run.go.test.-failfast.-timeout.20m.txt)

Maybe you can have a look if you happen to have time? But I think this is not related to this PR anyway...

PatrLind:
Ok, I have dived deep in the rabbit hole...
It turns out that the issue is caused by the use of `time.Now()` for telling when the config has changed in viper (https://github.com/ory/viper/blob/8cc0271ae7b63f6a696a9e586de4250dca19f33c/viper.go#L199).
This is in turn causing a problem with cache invalidation in the oathkeeper viper provider.
https://github.com/ory/oathkeeper/blob/45d5914ec6030e34ea37a91596df9a5548914c87/driver/configuration/provider_viper.go#L277

The problem with using `time.Now()` is that the precision of the time value is different on different operating systems. On Windows (at least my machine) it seems to give increase the time value every 1 millisecond.
Also, I think there might be another issue when the system clock is adjusted (on any system) that the timestamp can go backwards or stay the same for some longer time.
Anyway this will cause issues for cached values and can potentially be a security issue as well if the configuration isn't applied as the user thinks it should be.

So to recap. The reason the tests fail on Windows is because the old configuration is cached. This happens randomly depending on the timing of the tests. It will probably fail more on a fast computer than a slow computer.

So what is the solution? I am not exactly sure, but I think a good way to start would be to not use the configChangedAt in viper for anything related to caching or to tell if the configuration has changed. It would be better to use a configuration version number that increases monotonically for each change instead.
Also the `hashPipelineConfig` function scares me a bit since it creates a hash value and uses that for caching configuration. Seems to me that this should be a more cryptographically safe hash function instead.
https://github.com/ory/oathkeeper/blob/45d5914ec6030e34ea37a91596df9a5548914c87/driver/configuration/provider_viper.go#L293

aeneasr:
@PatrLind I think you have found https://github.com/ory/x/issues/216

zepatrik:
Thanks for your debugging effort :+1: 
Should we merge this PR as is? The failing cases in the CI definitely seem to be viper related.

PatrLind:
I guess you can merge, but no rush if you want to do more validation.

aeneasr:
Ah, my bad - I still think that this is not what is causing the issue (even though it definitely is an issue!) as the problem with the configs not being properly read in also appears in hydra, which does not use this method or value. In any case, I will try to debug this in ORY Hydra now and then will probably come back to the windows tests here!

Issue #153, opened 2019-01-25
Oathkeeper behind ssl terminating balancer


Entrio:
I have deployed ory os on various systems and I always run into the same issue when dealing with oathkeeper. In my case there is an nginx server that terminates all https requests and forwards the traffic to an oathkeeper upstream server.

Due to the nature of this setup, I always have to specify my match rules as http protocol as that is what nginx passes to oathkeeper. 

Can someone point me in the right direction on solving this issue? I could not find anything in the documentation.

Thanks

aeneasr:
Does the forwarded request include `X-Forwarded-Proto: https` in the header?

Entrio:
Yes, I have `X-Forwarded-Proto` set to `$scheme` (which in this case is https)

aeneasr:
No, in that case it would make sense to have `https` as the protocl for oathkeeper to match against!

Entrio:
This is my nginx config:

`server {

        server_name             gateway.***;

        location / {
                proxy_set_header                        Host                    $host;
                proxy_set_header                        X-Forwarded-For         $remote_addr;
                proxy_set_header                        X-Forwarded-Proto       $scheme;
                proxy_set_header                        X-Real-IP               $remote_addr;
                proxy_pass                              http://oathkeeper-proxy;
        }

        access_log              /var/log/nginx/gateway.access.log main;
        error_log               /var/log/nginx/gateway.error.log;

        listen          443     ssl; # managed by Certbot
        ssl_certificate         /etc/letsencrypt/live/gateway.***/fullchain.pem; # managed by Certbot
        ssl_certificate_key     /etc/letsencrypt/live/gateway.***/privkey.pem; # managed by Certbot
        include                 /etc/letsencrypt/options-ssl-nginx.conf; # managed by Certbot
        ssl_dhparam             /etc/letsencrypt/ssl-dhparams.pem; # managed by Certbot


}
`

In this case `$scheme` that is set is `https`. I will test with explicit `https` rather than `$scheme` and let you know how it goes

Entrio:
In the log i see the following message (it is clear it is passing http)

`Feb 15 16:00:35 lb1 oathkeeper[26591]: time="2019-02-15T16:00:35+06:00" level=warning msg="Access request denied" access_url="http://gateway.***/status" error="Requested url does not match any rules" granted=false
`

rauanmayemir:
[...]

PR #638, opened 2021-02-13
feat: Add support for  X-Forwarded-Proto header

christian-roggia:
## Related issue

https://github.com/ory/oathkeeper/issues/153 @aeneasr @Entrio

## Proposed changes

Add basic support for the header `X-Forwarded-Proto` which is normally set by an edge proxy like Ambassador, Istio, Traefik, Envoy, etc.

The proposed changes affect **only the decision API** as it is expected to be contacted by another proxy which will take care of SSL termination in most cases. It is important to notice that even if traffic reaching oathkeeper is plain HTTP, it does not mean that the traffic is not encrypted in transit. A good example is service mesh where Envoy or similar technologies (e.g. Linkerd) will secure traffic in transit through sidecars and mTLS.

## Checklist

      the changes.
      works.

## Further comments

**The changes proposed here have been tested in our development environment and they work like a charm for both proper rules matching and redirection URLs.**

### DIFF
--- a/api/decision.go
+++ b/api/decision.go
@@ -55,7 +55,7 @@ func (h *DecisionHandler) ServeHTTP(w http.ResponseWriter, r *http.Request, next
 	if len(r.URL.Path) >= len(DecisionPath) && r.URL.Path[:len(DecisionPath)] == DecisionPath {
 		r.URL.Scheme = "http"
 		r.URL.Host = r.Host
-		if r.TLS != nil {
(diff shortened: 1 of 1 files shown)


Entrio:
Awesome! Are you sure it was as simple as just that one line? @aeneasr comments?

aeneasr (approved):
Nice, LGTM!

PR #674, opened 2021-03-22
feat: Add health event manager and rules readiness probe

ArthurKnoep:
## Related issue

None

## Proposed changes

This PR adds an health event manager to update the readiness probe of oathkeeper.
It also includes a readiness probe for rules load (until the rules are loaded for the first time, the readiness endpoint will return an error).

This changes is mostly for usage with Kubernetes. Before we were observing a spike of 404 request on scale up. After some investigation it appears that it was the new instances of Oathkeeper that were added to the loadbalancer before the rules were fetched for the first time. After few tests on our staging cluster, we now have 0 errors on scale up.

## Checklist

      the changes.
      works.

## Further comments

While adding unit testing to the health event manager, I've been forced to use `time.Sleep` as I need some background task to complete (setting the watcher and dispatching and event). I'm not sure how I can improve this but I'm open to any better ideas.


driver/health/event_manager_default_test.go
```diff
@@ -0,0 +1,49 @@
...
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	t.Run("health event manager", func(t *testing.T) {
+		ruleReadinessProbe := rulereadiness.NewReadinessHealthChecker()
+		hem := NewDefaultHealthEventManager()
+
+		if err := hem.AddListener(ruleReadinessProbe); err != nil {
+			t.Errorf("AddListener() error = %v", err)
+			return
+		}
```
zepatrik, line 21: We use https://github.com/stretchr/testify usually for assertions:
```suggestion
		require.NoError(t, hem.AddListener(ruleReadinessProbe))
```

driver/health/event_manager_default.go
```diff
@@ -0,0 +1,51 @@
...
+	return &DefaultHealthEventManager{
+		evtChan:                make(chan interface{}),
+		listenerEventTypeCache: make(map[reflect.Type]Readiness),
+	}
+}
+
+func (h *DefaultHealthEventManager) Dispatch(event interface{}) {
+	select {
+	case h.evtChan <- event:
+	default:
```
zepatrik, line 24: Isn't this an error? Maybe return an error if the channel is blocked. Alternatively maybe spawn a go routine to dispatch the event later.

driver/health/event_manager_default.go
```diff
@@ -0,0 +1,51 @@
+package health
+
+import (
+	"context"
+	"errors"
```
zepatrik, line 5:
```suggestion
	"github.com/pkg/errors"
```

driver/health/event_manager_default.go
```diff
@@ -0,0 +1,51 @@
...
+	case h.evtChan <- event:
+	default:
+	}
+}
+
+func (h *DefaultHealthEventManager) AddListener(listener Readiness) error {
+	for _, evtType := range listener.EventTypes() {
+		evtTypeVal := reflect.TypeOf(evtType)
+		if _, ok := h.listenerEventTypeCache[evtTypeVal]; ok {
+			return errors.New("event type already registered")
```
zepatrik, line 32: It is better to export the error and return 
```suggestion
			return errors.WithStack(ErrEventTypeAlreadyRegistered)
```

driver/health/event_manager_default.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func (h *DefaultHealthEventManager) Dispatch(event interface{}) {
+	select {
+	case h.evtChan <- event:
+	default:
+	}
+}
+
+func (h *DefaultHealthEventManager) AddListener(listener Readiness) error {
```
zepatrik, line 28: This whole function is meant to only be called on initialization, right? Might be better to make this a `New(listeners ...Readines) (*DefaultHealthEventManager, error)` function.

driver/registry_memory.go
```diff
@@ -138,9 +145,30 @@ func (r *RegistryMemory) CredentialHandler() *api.CredentialsHandler {
...
+		healthCheckers := []health.Readiness{rulesReadinessChecker}
+
+		handlers := healthx.ReadyCheckers{}
+		for _, checker := range healthCheckers {
+			handlers[checker.Name()] = checker.Validate
+			if err := r.HealthEventManager().AddListener(checker); err != nil {
+				r.Logger().WithError(err).Fatalf("Unable to register health checker listener.")
+			}
+		}
+		r.healthxHandler = healthx.NewHandler(r.Writer(), r.BuildVersion(), handlers)
```
zepatrik, line 171: I would rather add a new function that returns a `healthx.ReadyCheckers`, so that you can unit test it and remove some logic from the registry.
```suggestion
		handlers, err := r.HealthEventManager().HealthxReadyCheckers()
		if err != nil {
			r.Logger().WithError(err).Fatalf("Unable to register health checker listener.")
		}
		r.healthxHandler = healthx.NewHandler(r.Writer(), r.BuildVersion(), handlers)
```

driver/health/event_manager_default_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+		if err := hem.AddListener(ruleReadinessProbe); err == nil {
+			t.Errorf("AddListener() was able to register twice the same listener")
+			return
+		}
```
zepatrik, line 26:
```suggestion
		assert.True(t, errors.Is(hem.AddListener(ruleReadinessProbe), ErrEventTypeAlreadyRegistered))
```

driver/health/event_manager_default_test.go
```diff
@@ -0,0 +1,49 @@
...
+		time.Sleep(100 * time.Millisecond)
+		// Dispatch event
+		hem.Dispatch(&rulereadiness.RuleLoadedEvent{})
+		// Wait for event propagation
+		time.Sleep(100 * time.Millisecond)
+
+		if err := ruleReadinessProbe.Validate(); err != nil {
+			t.Errorf("Validate() returned an unexpected error, err = %v", err)
+			return
+		}
```
zepatrik, line 46:
```suggestion
		require.NoError(t, ruleReadinessProbe.Validate())
```

driver/registry_memory.go
```diff
@@ -70,6 +74,9 @@ func (r *RegistryMemory) Init() {
 			r.Logger().WithError(err).Fatal("Access rule watcher terminated with an error.")
 		}
 	}()
+	go func() {
+		r.HealthEventManager().Watch(context.Background())
+	}()
```
zepatrik, line 79:
```suggestion
	r.HealthEventManager().Watch(context.Background())
```

driver/health/event_manager_default.go
```diff
@@ -0,0 +1,51 @@
...
+		evtTypeVal := reflect.TypeOf(evtType)
+		if _, ok := h.listenerEventTypeCache[evtTypeVal]; ok {
+			return errors.New("event type already registered")
+		}
+		h.listenerEventTypeCache[reflect.TypeOf(evtType)] = listener
+	}
+	return nil
+}
+
+func (h *DefaultHealthEventManager) Watch(ctx context.Context) {
```
zepatrik, line 39: This will always be called in a new go routine, so better just spawn it inside this function.

zepatrik (changes requested):
Nice :tada: 
Some quick feedback just to hint you a bit towards our code style.

driver/health/event_manager_default.go
ArthurKnoep: I've done the changes but now I don't know how to handle the error return on the constructor since it's used in the function `HealthEventManager` from registry_memory.go that does not return an error (and should not I think).

What I have done for now is to log the error and return nil, which will probably make the code panic. However, as an error can only be generated with a wrong inputs on the constructor and since the inputs provided to the constructor is statically set in the code, that should never happen.

That being said, I'm not really happy with that, what's your thoughts ?

driver/registry_memory.go
```diff
@@ -138,9 +143,23 @@ func (r *RegistryMemory) CredentialHandler() *api.CredentialsHandler {
 	return r.ch
 }
 
+func (r *RegistryMemory) HealthEventManager() health.EventManager {
+	if r.healthEventManager == nil {
+		var err error
+		rulesReadinessChecker := rulereadiness.NewReadinessHealthChecker()
+		if r.healthEventManager, err = health.NewDefaultHealthEventManager(rulesReadinessChecker); err != nil {
+			r.logger.WithError(err).Error("unable to instantiate new health event manager")
```
zepatrik, line 151: I think it is fine to fatal at this point and exit. If there is an error it would always happen on initialization right?
```suggestion
			r.logger.WithError(err).Fatal("unable to instantiate new health event manager")
```

rule/readiness/rule_readiness.go
```diff
@@ -0,0 +1,43 @@
...
+var ErrRuleNotYetLoaded = errors.New("rules have not been loaded yet")
+
+func NewReadinessHealthChecker() *RuleReadinessChecker {
+	return &RuleReadinessChecker{
+		hasFirstRuleLoad: false,
+	}
+}
+
+func (r *RuleReadinessChecker) Name() string {
+	return "rule-first-load"
```
aeneasr, line 24: Maybe make this a constant?

rule/readiness/rule_readiness_test.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+func TestReadinessHealthChecker(t *testing.T) {
+	t.Run("rule readiness probe", func(t *testing.T) {
+		ruleReadinessProbe := NewReadinessHealthChecker()
+
+		if name := ruleReadinessProbe.Name(); name != "rule-first-load" {
+			t.Errorf("Name() did not returned expected name, name = %s", name)
+			return
+		}
```
zepatrik, line 14: Also these tests should use `require` and `assert` :wink: 
```suggestion
		assert.Equal(t, "rule-first-load", ruleReadinessProbe.Name())
```

rule/readiness/rule_readiness_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+		if err := ruleReadinessProbe.Validate(); err != ErrRuleNotYetLoaded {
+			t.Errorf("Validate() did not returned expected error, error = %v", err)
+			return
+		}
```
zepatrik, line 19:
```suggestion
		assert.True(t, errors.Is(ruleReadinessProbe.Validate(), ErrRuleNotYetLoaded)
```

rule/readiness/rule_readiness_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+		evtTypes := ruleReadinessProbe.EventTypes()
+		if len(evtTypes) != 1 {
+			t.Errorf("EventTypes() returned either 0 or multiple event type, evtTypes = %v", evtTypes)
+			return
+		}
```
zepatrik, line 25:
```suggestion
		require.Len(t, evtTypes, 1)
```

driver/health/event_manager_default.go
```diff
@@ -0,0 +1,63 @@
...
+	var listenerEventTypeCache = make(map[reflect.Type]Readiness)
+	for _, listener := range listeners {
+		for _, evtType := range listener.EventTypes() {
+			evtTypeVal := reflect.TypeOf(evtType)
+			if _, ok := listenerEventTypeCache[evtTypeVal]; ok {
+				return nil, errors.WithStack(ErrEventTypeAlreadyRegistered)
+			}
+			listenerEventTypeCache[reflect.TypeOf(evtType)] = listener
+		}
+	}
```
aeneasr, line 27: I've never actually used reflection so I don't really understand what's going on here. Maybe you could explain the idea behind this? :)

rule/readiness/rule_readiness_test.go
```diff
@@ -0,0 +1,39 @@
...
+		if _, ok := evtTypes[0].(*RuleLoadedEvent); !ok {
+			t.Errorf("EventTypes() returned an unkown event type, evtTypes = %v", evtTypes)
+			return
+		}
```
zepatrik, line 29:
```suggestion
		_, ok := evtTypes[0].(*RuleLoadedEvent)
		assert.True(t, ok, "actual type %T", evtTypes[0])
```

rule/readiness/rule_readiness_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+		// Dispatch fake event
+		ruleReadinessProbe.EventsReceiver(&RuleLoadedEvent{})
+
+		if err := ruleReadinessProbe.Validate(); err != nil {
+			t.Errorf("Validate() returned an error, error = %v", err)
+			return
+		}
```
zepatrik, line 37:
```suggestion
		assert.NoError(t, ruleReadinessProbe.Validate())
```

rule/readiness/rule_readiness.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+func (r *RuleReadinessChecker) Validate() error {
+	if !r.hasFirstRuleLoad {
+		return ErrRuleNotYetLoaded
```
zepatrik, line 29:
```suggestion
		return errors.WithStack(ErrRuleNotYetLoaded)
```

aeneasr (commented):
Thank you and 🥳 for your first contribution!

So for understanding a bit what's going on - we have a generic event manager that we can send events to. And the event manager decides if we're live or not?

Then we have events for rules which returns an error on Validate if it's not yet ready?

driver/health/event_manager_default_test.go
```diff
@@ -0,0 +1,54 @@
...
+		// Rule readiness probe must return an error before event dispatch
+		require.Error(t, ruleReadinessProbe.Validate())
+
+		// Dispatch event without watching (should not block)
+		hem.Dispatch(&rulereadiness.RuleLoadedEvent{})
+
+		// Watching for incoming events
+		go func() {
+			hem.Watch(ctx)
+		}()
```
zepatrik, line 42:
```suggestion
		hem.Watch(ctx)
```

zepatrik (changes requested):
Nice, this looks pretty good now. What I am interested in is how does this handle reloading of rules? I am not too deep into the Oathkeeper code base right now, maybe you already thought about that? I assume there is no downtime on reloads and it just takes a few milliseconds to load everything? So the health handlers probably don't have to reflect that.

driver/registry_memory.go
ArthurKnoep: Yes exactly

rule/readiness/rule_readiness_test.go
ArthurKnoep: yep forgot this file sorry

ArthurKnoep:
> Nice, this looks pretty good now. What I am interested in is how does this handle reloading of rules? I am not too deep into the Oathkeeper code base right now, maybe you already thought about that? I assume there is no downtime on reloads and it just takes a few milliseconds to load everything? So the health handlers probably don't have to reflect that.

From what I have observed rules reload does not create any down time as it simply overwrite the rules saved in the registry. The state of my rule readiness will also not change since event if I re-dispatch a `RuleLoadedEvent`, the event consumer will simply sets again the boolean to true. I know that I will be using some useless resources to dispatch the event and consume it but since everything is done asynchonously and efficiently I think that's ok.

ArthurKnoep:
> Thank you and  for your first contribution!
> 
> So for understanding a bit what's going on - we have a generic event manager that we can send events to. And the event manager decides if we're live or not?
> 
> Then we have events for rules which returns an error on Validate if it's not yet ready?

Yes exactly, while implementing my readiness probe for the rule load I noticed that the `/health/alive` was not really implemented (I mean it was just an empty `healthx.HealthCheckers`, which would have always returned "ok"). So I've decided to implement a generic health event manager so that if in the future we need to add other probes elsewhere in the code we would be able to re-use a single instance of this manger instead of a multitude of different probe instance.
The rule load readiness probe I've implemented returns an health error until the rules are firstly loaded (which have for goal to avoid oathkeeper being added to the load balancing pool too early)

While implementing the event distpatcher I hesitated with two different architectures. 
The first one was to deliver the events to all the listeners and it was up to the listeners to decide whether they wanted to consume the event or not (using a select...default mechanism)
The second one was using `reflect.TypeOf` to determine the type of the event received and to dispatch it directly to the correct listener.

I've decided to go with the second mechanism as it allows me to save in a map an association between an event type and a listener, that way when I receive an event I can efficiently dispatch it to the correct listener.

driver/health/event_manager_default.go
ArthurKnoep: The idea behind those few lines is to generate a map that associate an event type (aka a struct type) to a listener. 
The general idea is that `reflect.TypeOf` returns a `reflect.Type` that can be compared to another one to ensure that two different variables have the same type.

So for that, I iterate through each listeners I receive in argument, for each listener I call the `EventTypes` method which returns me a slice of interface that contains an instance of each event struct the probe needs to watch.
In the case of my rule readiness probe, I need to watch `readiness.RuleLoadedEvent` so the `EventTypes` function looks like something like this: 
```go
return []interface{}{&RuleLoadedEvent}
```
Then for each event types, I compute the `reflect.Type` value that I store in a map with the listener.

So now, when I receive an event to dispatch I simply have to compute its "type" using reflection, to check if I have something in my map, and if it's the case, to call the `EventsReceiver` function.

The reason of my choice for reflection is in my message below: https://github.com/ory/oathkeeper/pull/674#issuecomment-805292010

zepatrik:
Thanks for the explanation. I like the approach, especially as I looked into a similar case recently. I tried registering callbacks and polling, which was quite messy and did not work nicely. This looks way nicer, and I guess we can implement it in other projects as well, e.g. https://github.com/ory/keto/issues/472

driver/health/event_manager_default.go
```diff
@@ -0,0 +1,63 @@
...
+import (
+	"context"
+	"reflect"
+
+	"github.com/ory/x/healthx"
+	"github.com/pkg/errors"
+)
+
+type DefaultHealthEventManager struct {
+	evtChan                chan interface{}
```
aeneasr, line 12: By making this an interface with `HealthListenerID() string` or optionally a global enum we can avoid all of the reflection. This would be more along the lines of Go's "idiomatic" design. The problem with reflection is one of prefixing (I have encountered extremely hard to debug issues in gobufallo/pop related to this) and using an interface here is much more obvious to the (code) observer. What do you think?

ArthurKnoep: By type name clashes do you mean something like that:
```
./root/
 |-> a/
 |   `-> test/test.go      // defines a struct named Test
 `-> b/
     `-> test/test.go      // defines a struct also named Test
```
Meaning their types would both be `test.Test`

If that's what you are refering, reflect type comparaison is safe for that, I've joined a small demonstration:
Capture d’écran de 2021-03-24 10-44-10

However your idea of using an event interface is indeed quite clearer and will avoid the obfuscation caused by reflection, I'll update the code.

aeneasr: Ah, that makes sense - I thought we're using string representation where this is a problem :)

> However your idea of using an event interface is indeed quite clearer and will avoid the obfuscation caused by reflection, I'll update the code.

Cool! :)

ArthurKnoep: I've done the changes :wink:

api/health_test.go
```diff
@@ -37,7 +37,7 @@ func TestHealth(t *testing.T) {
 	res, err = server.Client().Get(server.URL + "/health/ready")
 	require.NoError(t, err)
 	defer res.Body.Close()
-	require.Equal(t, http.StatusOK, res.StatusCode)
+	require.Equal(t, http.StatusServiceUnavailable, res.StatusCode)
 	require.NoError(t, json.NewDecoder(res.Body).Decode(&result))
 	assert.Equal(t, "ok", result.Status)
```
aeneasr, line 42: Shouldn't this not be "ok" if we get `http.StatusServiceUnavailable`?

driver/health/event_manager_default.go
```diff
@@ -0,0 +1,64 @@
...
+		for {
+			var evt ReadinessProbeEvent
+			select {
+			case evt = <-h.evtChan:
+			case <-ctx.Done():
+				return
+			}
+			if listener, ok := h.listenerEventTypeCache[evt.ReadinessProbeListenerID()]; ok {
+				listener.EventsReceiver(evt)
+			}
```
aeneasr, line 53: I think it would be cleaner and easier to understand if done like this:

```suggestion
			select {
			case <-ctx.Done():
				return
			}
			case evt, ok := <-h.evtChan:
				if !ok { // channel closed
					return
				}
				if listener, ok := h.listenerEventTypeCache[evt.ReadinessProbeListenerID()]; ok {
					listener.EventsReceiver(evt)
				}
```

aeneasr (commented):
Sorry for the long long long wait times. Sometimes things just get too hectic here and I forget to review these things.

The PR looks good, I have a minor nit pick and one question but I think these can be resolved fast and we should be good to go asap! Next review will be immediate, promised :)

api/health_test.go
ArthurKnoep: The thing is that at this point of the execution of the test the rules are not yet loaded, so that's why I'm expecting a service unavailable error.
However now that I think of it, I could probably wait loop with a timeout for this check to be `http.StatusOk`

aeneasr: Right but what I meant is that the body should not be `{"status":"ok"}` if the status is not ok (because the service is not available). Or am I missing something?

ArthurKnoep: Indeed that's weird, I'll check that

ArthurKnoep: So I've found why: the variable `result` was still containing the value of the previous test (`/health/alive`).

But in any case I've fix the test, now I init the registry created at the begining of the test file, and therefore I load the rules.
I've also added a sleep of 100ms to be sure that the event had the time to propagate, but thanks to the alive http query it was not necessary (I've still added it to be absolutely sure)

api/health_test.go
```diff
@@ -16,6 +17,10 @@ import (
 func TestHealth(t *testing.T) {
 	conf := internal.NewConfigurationWithDefaults()
 	r := internal.NewRegistry(conf)
+	r.Init()
+
+	// Waiting for rule load and health event propagation
```
aeneasr, line 22: Shouldn't we check if the service is unavailable before? That way we know if the ruleloader actually works!

ArthurKnoep: easy peasy :sunglasses:

aeneasr (approved):

PR #681, opened 2021-03-28
feat: additional JWT auth debug information

### LINKED ISSUE #668: Additional log information for denied requests(JWT auth)
**Is your feature request related to a problem? Please describe.**
When a company offers an authorized public API, information about which user is getting forbidden and/or invalid requests comes in very handy for various support issues.

**Describe the solution you'd like**
Append a configurable "identifier" to error logs outputted after JWT verification(from `pipeline/authn/authenticator_jwt.go` L98 on).

The suggested approach is to have a new configuration value within `AuthenticatorOAuth2JWTConfiguration` which would tell Oathkeeper to extract the value from "Claims"(similar to what `credentials/verifier_default.go` does for validation purposes) and log the "identifier key".
As an example, the JWT verifier could be configured to extract the key "email" from the decoded JWT and log the email as part of the log entry - the company would then be able to assist the client and verify their access configuration(the configuration on the JWT issuer).

**Describe alternatives you've considered**
Logging could be done one request level before(in our case Ambassador API gateway), but Oathkeeper seems a better location for such logs since the authorization happens within Oathkeeper.

**Additional context**
If the feature is interesting and the maintainers approve it for the inclusion, let me know and I'll prepare a PR.

eroznik:
JWT Claims added to error details field. Closes #668

## Related issue
#668 @aeneasr 

## Proposed changes
Added `jwt_claims` to the result error details field in cases where the JWT Authenticator failed to authorize the provided JWT.

## Checklist

      the changes.
      works.

### DIFF
--- a/pipeline/authn/authenticator_jwt.go
+++ b/pipeline/authn/authenticator_jwt.go
@@ -2,10 +2,11 @@ package authn
 
 import (
 	"encoding/json"
-	"net/http"
-
+	"fmt"
 	"github.com/form3tech-oss/jwt-go"
 	"github.com/pkg/errors"
+	"net/http"
+	"strings"
 
 	"github.com/ory/go-convenience/jwtx"
 	"github.com/ory/herodot"
@@ -96,7 +97,7 @@ func (a *AuthenticatorJWT) Authenticate(r *http.Request, session *Authentication
 		ScopeStrategy: a.c.ToScopeStrategy(cf.ScopeStrategy, "authenticators.jwt.Config.scope_strategy"),
 	})
 	if err != nil {
-		return helper.ErrUnauthorized.WithReason(err.Error()).WithTrace(err)
+		return a.tryEnrichResultErr(token, helper.ErrUnauthorized.WithReason(err.Error()).WithTrace(err))
 	}
 
 	claims, ok := pt.Claims.(jwt.MapClaims)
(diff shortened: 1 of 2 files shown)


pipeline/authn/authenticator_jwt.go
```diff
@@ -109,3 +111,29 @@ func (a *AuthenticatorJWT) Authenticate(r *http.Request, session *Authentication
...
+	if !ok {
+		return err
+	}
+	var claimKeyPairs []string
+	for k, v := range claims {
+		// print just root content of the JWT, skip nested objects
+		if _, ok := v.(map[string]interface{}); ok {
+			continue
+		}
+		if floatVal, ok := v.(float64); ok && v == float64(int64(floatVal)) {
```
aeneasr, line 130: This checks if v is a float, then converts it to an int, and then back to a float, and then back to an int. Do we really need this? Why no tjust `fmt.Sprintf("%v", v)`?

eroznik: this isn't really the pretties thing, but i wanted to print the full integer value and not something like this `iat=1.616678102e+09`(this is the result using `fmt.Sprintf("%v", v)`)

aeneasr: The easiest way to do that is to format the output as JSON then, so:

```
var b bytes.Buffer
json.Encode(&b, claims)
err.WithDetail("claims", b.String())
```

which would also make it easier to process in log systems because it is a standard format

aeneasr (approved):

PR #688, opened 2021-04-06
chore: add prettier

zepatrik:
Also added doctoc script and ran it.


aeneasr:
tests fail

zepatrik:
Fixed it, there seems to be a breaking change in the SDK for some reason....

PR #694, opened 2021-04-20
fix: update shebangs to use bash from /usr/bin/env instead of /bin/ for better portability

### LINKED ISSUE #693: Unable to build oathkeeper from source on systems where bash is not in /bin/
**Describe the bug**

On some systems, bash doesn't lie in `/bin/`, but in other places, like `/usr/local/bin/bash` (FreeBSD) or even in unpredictable paths like `/nix/store/y0vp7qcjz02rsgawlgmx8ilbh2wimcg5-system-path/bin/bash` (NixOS). It makes the installation impossible without modifying the Makefile.

**Reproducing the bug**

1. Try to build oathkeeper from source on a system where bash is not in `/bin/`
2. Command `GO111MODULE=on make install-stable` fails with the following output:
```
make: /bin/bash: No such file or directory
make: /bin/bash: No such file or directory
go build -o .bin/go-acc github.com/ory/go-acc
make: /bin/bash: No such file or directory
make: *** [Makefile:9: deps] Error 127
```

**Expected behavior**

Build fails without errors.

**Environment**

- Version: e44405948dd32ac70435ee9199ed5d1f85de16fb
- Environment: NixOS

**Additional context**

I'd suggest using `/usr/bin/env bash` instead of `/bin/bash` since env is more likely to be in `/usr/bin` than bash is to be in `/bin`.

MrMebelMan:
Fixes #693

## Proposed changes

As said in the issue, on some systems, bash doesn't lie in /bin/, but in other places, like /usr/local/bin/bash (FreeBSD) or even in unpredictable paths like /nix/store/y0vp7qcjz02rsgawlgmx8ilbh2wimcg5-system-path/bin/bash (NixOS).
It makes the installation impossible without modifying the Makefile.
This PR replaces all `/bin/bash` shebangs with `/usr/bin/env bash`, improving portability.

## Checklist

      the changes.
      works.

## Further comments

I consider `/usr/bin env` to be more portable, because env is more likely to be in `/usr/bin` than bash is to be in `/bin`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-SHELL=/bin/bash -o pipefail
+SHELL=/usr/bin/env bash -o pipefail
 
 export GO111MODULE := on
 export PATH := .bin:${PATH}
--- a/scripts/run-format.sh
+++ b/scripts/run-format.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -euo pipefail
 
--- a/scripts/run-gensdk.sh
+++ b/scripts/run-gensdk.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -euo pipefail
 
--- a/scripts/run-genswag.sh
+++ b/scripts/run-genswag.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -euo pipefail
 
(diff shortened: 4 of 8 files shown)


MrMebelMan:
Lol, since the CI tests have failed, I suppose `/usr/bin env` is not as portable as I thought 😅

aeneasr (approved):
😅 the CI failed because of a different reason but if you run `make format` it should pass!

aeneasr:
I think this is good to merge as is

PR #695, opened 2021-04-22
docs: specify public keys only

stensonb:
the .well-known/jwks.json endpoint should (and does) only return the public keys of the JWKS (as described here: https://www.ory.sh/oathkeeper/docs/pipeline/mutator#configuration-1)

## Proposed changes

clarifying the data delivered at the .well-known/jwks.json endoint

## Checklist

      the changes.
      works.

### DIFF
--- a/docs/docs/reference/api.mdx
+++ b/docs/docs/reference/api.mdx
@@ -40,7 +40,7 @@ Accept: application/json
 
 ```
 
-This endpoint returns cryptographic keys that are required to, for example,
+This endpoint returns cryptographic public keys that are required to, for example,
 verify signatures of ID Tokens.
 
 #### Responses
@@ -78,22 +78,11 @@ Status Code **500**
   "keys": [
     {
       "alg": "string",
-      "crv": "string",
-      "d": "string",
-      "dp": "string",
-      "dq": "string",
       "e": "string",
-      "k": "string",
       "kid": "string",
       "kty": "string",
       "n": "string",
-      "p": "string",
-      "q": "string",
-      "qi": "string",
       "use": "string",
-      "x": "string",
-      "x5c": ["string"],
-      "y": "string"
     }
   ]
 }


aeneasr (changes requested):
Thank you!

Unfortunately as it says in the top of the file - this file is autogenerated and the source needs to be updated: https://github.com/ory/oathkeeper/blob/d942c043aa2370b87e0dd822440ad39d809755f9/api/credential.go#L48

stensonb:
yup, sure enough.  sorry, just getting the hang of the doc generator. :)

this does bring up a separate issue, however.  the generator indicates the `200` response is the entire `jsonWebKeySet`, which isn't accurate: https://github.com/ory/oathkeeper/blob/d942c043aa2370b87e0dd822440ad39d809755f9/api/credential.go#L56

the 200 response is actually a subset (only the public key details): https://github.com/ory/oathkeeper/blob/d942c043aa2370b87e0dd822440ad39d809755f9/api/credential.go#L70-L77

i'll create a private struct to hold the public key details and return that (and update the docs to reflect the private struct).  updated PR coming soon... :)

aeneasr:
> yup, sure enough. sorry, just getting the hang of the doc generator. :)

Don't worry - just thinking about how we can help contributors see this better as it's not the first time :) If you have any ideas please let me know!

> the 200 response is actually a subset (only the public key details):

So basically a different struct without the private key fields? I think that would be ok but is not really necessary in my opinion as noone is really parsing those fields themselves and they usually just pass them to some library. The fields are `omitempty` so basically not set usually! From my side no need to spend work on that :)

aeneasr:
Are you still up for the changes? :) If you need any help, let us know!

aeneasr:
Are you still up for the changes? :monocle_face: If you need any help, let us know!

I'll convert this to a draft in the meanwhile. When it's ready for review again, please remove the draft indicator! 👀

Issue #181, opened 2019-04-30
Trusting self-signed certificates


jboalml:
I am trying to connect to a resource server using Oathkeeper in Docker for Windows and, since I am still in my development setup, I am using self-signed certificates. So far, I have successfully managed to hit the resource server, but Oathkeeper keeps complaining about the server certificate being signed by an unknown authority. The log message is the following:

```
time="2019-04-18T15:43:38Z" level=warning msg="Access request denied because roundtrip failed" error="x509: certificate signed by unknown authority" granted=false
2019/04/18 15:43:38 http: proxy error: x509: certificate signed by unknown authority
```

It would be useful to support using self-signed certificates in Oathkeeper to allow testing HTTPS connections before going into production. I suggest two options:
* Incorporate a flag to disable resource server certificate verification, just like ``--skip-tls-verify`` in Hydra. However, this should not affect incoming TLS connections.
* Being able to feed the certificate chain to perform validation.

I believe both alternatives have different use cases. The first one would allow to quickly test HTTPS communication with resource servers, whereas the second would be useful to secure last mile connections in those scenarios where you cannot or do not want to use certificates signed by a CA.

Any help would be greatly appreciated.

P.S. This an extended repost of the following [forum message](https://community.ory.sh/t/trusting-self-signed-certificate-in-development/1081) suggested by @arekkas.

aeneasr:
Just to get it right - you want the connectivity to the upstream server to be encrypted in transport (HTTP over TLS) with self-signed certificate which fails because of CA verification?

jboalml:
> Just to get it right - you want the connectivity to the upstream server to be encrypted in transport (HTTP over TLS) with self-signed certificate which fails because of CA verification?

Yes, this is exactly the problem. My development resource server does not have a public domain name and therefore I cannot use a certificate signed by a CA. The only solution I have found so far is to secure Oathkeeper's incoming requests and change my upstream redirection to HTTP.

aeneasr:
Sorry for the late reply, sounds like a reasonable feature request. Open to accept PRs!

wraix:
I have created a function that should do what is required, but im having trouble figuring out how to implement the configuration part. Help would be appreciated. - At least this shows how to implement extending the Root CA for upstream connections.

This is not ready to be merged. I just did not understand how pull requests work on forks. Is my first time, sorry.

Zero-2:
Any news about this?

wraix:
[...]

PR #706, opened 2021-04-25
feat: Extended Root CA for upstream connections (#181)

wraix:
This allows for appending a certificate file to the Root CA without altering the system Root CA. This is useful for allowing self-signed certificates on the upstream connections

## Checklist

      the changes.
      works.


wraix:
I got a notification from github saying: "Windows go test #412 - No test was run". Not sure what that means?

aeneasr (commented):
Thank you! This looks great! Could you please add a test for this? I think you could use `httptest.NewTLSServer` and extract the certificate from it

wraix:
I have added a test case to test that appending a certificate file to root ca works for TLS on upstream connections.

Enjoy.

aeneasr:
Thank you! Hectic week...will review on Monday!

driver/registry_memory.go
```diff
@@ -102,6 +106,41 @@ func (r *RegistryMemory) RuleMatcher() rule.Matcher {
 	return r.ruleRepository
 }
 
+func (r *RegistryMemory) UpstreamTransport(req *http.Request) (http.RoundTripper, error) {
+
+	// Use req to decide the transport per request iff need be.
+
+	certFile := r.c.ProxyServeUpstreamCaAppendCrtPath()
```
aeneasr, line 113: It looks like this would be called on every `UpstreamTransport` which means reading the cert file on every single request. This will cause significant issues at scale. Instead, we should cache the TLS certification / transport. Keep in mind that the config may change at any point in time so it is probably a good idea to invalidate the cache if the `certFile` value has changed.

aeneasr (commented):
This looks really good. I think we need to improve the TLS set up so that it doesn't load on every request. Also it would be great if you could document this new feature [as a new section `Upstream Configuration`](https://github.com/ory/oathkeeper/blob/master/docs/docs/api-access-rules.md#handler-configuration).

Thanks!

aeneasr:
Are you still up for the changes? :monocle_face: If you need any help, let us know!

christian-roggia:
+1 this is extremely useful for internal load balancers that communicate over HTTPS.

wraix:
Yes i will try and find some time for this. Might need some guidance on where to put the cached transport, but will give it a go.

driver/registry_memory.go
aeneasr: @wraix you can put the cached transport in a var that belongs to `RegistryMemory` so something like:

```
type RegistryMemory {
  upstreamTransport *http.Transport
  cachedCertFilePath string
}
```

and use `cachedCertFilePath` to check if the file location changed so that you can invalidate the cache. Alternatively you could also `os.Stat` the file to check if its mod time was changed, but this would again require disk access which is expensive. As an alternative, you could make the `os.Stat` check every 1 / 1000 requests (or something similar) giving a decent trade off between cache staleness and speed.

wraix: Might do both and put some configuration for the os.Stat check value.

Thanks for the pointer - will give it a Go.

wraix: @aeneasr  I made an implementation. Not sure if test cases for the certificate refresh is needed or how to perform those in a test suite.

driver/registry_memory.go
```diff
@@ -101,6 +111,108 @@ func (r *RegistryMemory) RuleMatcher() rule.Matcher {
...
+		RootCAs:            rootCAs,
+	}
+
+	return transport, &stat, nil
+}
+
+// UpstreamTransport decides the transport to use for the upstream for the request.
+func (r *RegistryMemory) UpstreamTransport(req *http.Request) (http.RoundTripper, error) {
+
+	r.upstreamRequestCount++
```
aeneasr, line 185: Since we probably have highly concurrent system this might cause issues. I think it would make sense to use atomic.Uint instead ( https://golang.org/pkg/sync/atomic/ ) to count up here :)

aeneasr (changes requested):
Awesome, thank you! This looks much better. Just one more thing :)

aeneasr:
@christian-roggia is continuing the work done here in #744 :) So I'll be marking this as a draft!

aeneasr:
Closing this as it is superseded by #744

PR #717, opened 2021-05-04
feat: Ability to Configure Remote Authorizers to set Headers in AuthenticationSession

wesleyfantinel:
## Proposed changes

The remote authorizers may have useful context from user's permissions. So with this changes, custom authorizers using `remote` and `remote_json` can return some useful headers to be forward into the AuthenticationSession, meaning that these headers will be passed to upstream services. 

For example, an user containing scopes/branches inside an organization profile has some level of data addressed to him. In this case, the upstream service need to know that, and "filter" the data according to his "branch_id". The permission that is given to the user (and the remote authorizers manages) has a record of the "branch_id", for the following responses will be returned as status code 200 (if granted) and containing a header like `X-Branch-Id`. 

The upstream service receives the `X-Branch-Id` and does your thing.

**The configuration requires to configure a list of "allowed headers" returning from remote authorizer, that will be accepted in the pipeline.**

---

Without this, we have to perform another request using the mutator `Hydrador` to retrieve the users' "branch_id", that is not a good performance approach in my vision for this set. 

## Checklist
      the changes.
      works.

### DIFF
--- a/.schema/config.schema.json
+++ b/.schema/config.schema.json
@@ -929,6 +929,17 @@
           "additionalProperties": {
             "type": "string"
           }
+        },
+        "allowed_remote_headers": {
+          "description": "A list of non simple headers the remote is allowed to return to mutate requests.",
+          "title": "Allowed Remote HTTP Headers for his Responses",
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "minLength": 0,
+          "uniqueItems": true,
+          "default": []
         }
       },
       "required": [
@@ -957,6 +968,17 @@
           "examples": [
             "{\"subject\":\"{{ .Subject }}\"}"
           ]
+        },
+        "allowed_remote_headers": {
+          "description": "A list of non simple headers the remote is allowed to return to mutate requests.",
+          "title": "Allowed Remote HTTP Headers for his Responses",
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "minLength": 0,
+          "uniqueItems": true,
+          "default": []
         }
       },
       "required": [
--- a/.schema/config.schema.json
+++ b/.schema/config.schema.json
@@ -930,7 +930,7 @@
             "type": "string"
           }
         },
-        "allowed_remote_headers": {
+        "forward_response_headers_to_upstream": {
           "description": "A list of non simple headers the remote is allowed to return to mutate requests.",
           "title": "Allowed Remote HTTP Headers for his Responses",
           "type": "array",
@@ -969,7 +969,7 @@
             "{\"subject\":\"{{ .Subject }}\"}"
           ]
         },
-        "allowed_remote_headers": {
+        "forward_response_headers_to_upstream": {
           "description": "A list of non simple headers the remote is allowed to return to mutate requests.",
           "title": "Allowed Remote HTTP Headers for his Responses",
           "type": "array",
--- a/.schemas/authorizers.remote_json.schema.json
+++ b/.schemas/authorizers.remote_json.schema.json
@@ -30,7 +30,7 @@
       },
       "description": "A list of non simple headers the remote is allowed to return to mutate requests.",
       "examples": [
-        "X-Branch-Id"
+        "X-Foo"
       ]
     }
   },
--- a/.schemas/authorizers.remote_json.schema.json
+++ b/.schemas/authorizers.remote_json.schema.json
@@ -22,7 +22,7 @@
         "{\"subject\":\"{{ .Subject }}\"}"
       ]
     },
-    "allowed_remote_headers": {
+        "forward_response_headers_to_upstream": {
       "title": "Allowed Remote HTTP Headers for his Responses",
       "type": "array",
       "items": {
--- a/.schemas/authorizers.remote_json.schema.json
+++ b/.schemas/authorizers.remote_json.schema.json
@@ -21,6 +21,17 @@
       "examples": [
         "{\"subject\":\"{{ .Subject }}\"}"
       ]
+    },
+    "allowed_remote_headers": {
+      "title": "Allowed Remote HTTP Headers for his Responses",
+      "type": "array",
+      "items": {
+        "type": "string"
+      },
+      "description": "A list of non simple headers the remote is allowed to return to mutate requests.",
+      "examples": [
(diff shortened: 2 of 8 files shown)


aeneasr (changes requested):
Thank you for the contribution! Unfortunately we can not accept contributions without tests. These are important to ensure that your feature does what it should, and it ensures that it keeps working even if others change your code!

if you need help with the tests feel free to ping a maintainer!

wesleyfantinel:
Ok, sending the tests.

aeneasr:
Looks like tests are failing for the new feature, could you take a look maybe? 🧐

docs/docs/pipeline/authz.md
```diff
@@ -344,6 +347,9 @@ authorizers:
       "remote": "http://my-remote-authorizer/authorize",
       "headers": {
         "X-Subject": "{{ print .Subject }}"
+      },
+      "allowed_remote_headers": {
+        "X-Branch-Id"
       }
```
aeneasr, line 353: Shouldn't this be an array?

```suggestion
      "allowed_remote_headers": [
        "X-Branch-Id"
      ]
```

.schema/config.schema.json
```diff
@@ -929,6 +929,17 @@
           "additionalProperties": {
             "type": "string"
           }
+        },
+        "allowed_remote_headers": {
```
aeneasr, line 933: I think this should have a different name to more clearly indicate what it is doing. What do you think about:

```suggestion
        "forward_response_headers_to_upstream": {
```

.schema/config.schema.json
```diff
@@ -957,6 +968,17 @@
           "examples": [
             "{\"subject\":\"{{ .Subject }}\"}"
           ]
+        },
+        "allowed_remote_headers": {
```
aeneasr, line 972: I think this should have a different name to more clearly indicate what it is doing. What do you think about:

```suggestion
        "forward_response_headers_to_upstream": {
```

.schemas/authorizers.remote_json.schema.json
```diff
@@ -21,6 +21,17 @@
       "examples": [
         "{\"subject\":\"{{ .Subject }}\"}"
       ]
+    },
+    "allowed_remote_headers": {
```
aeneasr, line 25: I think this should have a different name to more clearly indicate what it is doing. What do you think about:

```suggestion
        "forward_response_headers_to_upstream": {
```

docs/docs/pipeline/authz.md
```diff
@@ -292,6 +292,9 @@ if it returns a "403 Forbidden" response code, the access is denied.
   to an
   [`AuthenticationSession`](https://github.com/ory/oathkeeper/blob/master/pipeline/authn/authenticator.go#L40)
   object. See [Session](../pipeline.md#session) for more details.
+- `allowed_remote_headers` (slice of strings, optional) - The HTTP headers that
```
aeneasr, line 295:
```suggestion
- `forward_response_headers_to_upstream` (slice of strings, optional) - The HTTP headers that 
```

docs/docs/pipeline/authz.md
```diff
@@ -430,6 +440,9 @@ authorizers:
     "config": {
       "remote": "http://my-remote-authorizer/authorize",
       "payload": "{\"subject\": \"{{ print .Subject }}\", \"resource\": \"{{ printIndex .MatchContext.RegexpCaptureGroups 0 }}\"}"
+    },
+    "allowed_remote_headers": {
+      "X-Branch-Id"
     }
```
aeneasr, line 446: Should also be an array and renamed

docs/docs/pipeline/authz.md
```diff
@@ -373,6 +379,10 @@ Forbidden" response code, the access is denied.
   to an
   [`AuthenticationSession`](https://github.com/ory/oathkeeper/blob/master/pipeline/authn/authenticator.go#L40)
   object. See [Session](../pipeline.md#session) for more details.
+- `allowed_remote_headers` (slice of strings, optional) - The HTTP headers that
```
aeneasr, line 382: Should also be renamed

internal/config/.oathkeeper.yaml
```diff
@@ -264,6 +264,8 @@ authorizers:
     config:
       remote: https://host/path
       headers: {}
+      allowed_remote_headers:
```
aeneasr, line 267: Don't forget to update this one :)

internal/config/.oathkeeper.yaml
```diff
@@ -273,6 +275,8 @@ authorizers:
     config:
       remote: https://host/path
       payload: "{}"
+      allowed_remote_headers:
```
aeneasr, line 278: Don't forget to update this one :)

pipeline/authz/remote.go
```diff
@@ -21,8 +21,9 @@ import (
 
 // AuthorizerRemoteConfiguration represents a configuration for the remote authorizer.
 type AuthorizerRemoteConfiguration struct {
-	Remote  string            `json:"remote"`
-	Headers map[string]string `json:"headers"`
+	Remote               string            `json:"remote"`
+	Headers              map[string]string `json:"headers"`
+	AllowedRemoteHeaders []string          `json:"allowed_remote_headers"`
```
aeneasr, line 26: Don't forget to update this one :)

pipeline/authz/remote_json.go
```diff
@@ -21,8 +21,9 @@ import (
 
 // AuthorizerRemoteJSONConfiguration represents a configuration for the remote_json authorizer.
 type AuthorizerRemoteJSONConfiguration struct {
-	Remote  string `json:"remote"`
-	Payload string `json:"payload"`
+	Remote               string   `json:"remote"`
+	Payload              string   `json:"payload"`
+	AllowedRemoteHeaders []string `json:"allowed_remote_headers"`
```
aeneasr, line 26: Don't forget to update this one :)

aeneasr (approved):
This looks ready to be merged! The tests look great and the implementation too. I just would like to ask to rename the field to make it more obvious what it does :)

aeneasr (approved):
Great job, thank you for your hard work!

Issue #512, opened 2020-09-08
oathkeeper Authorizers  handler can not  receiving values from  the AuthenticationSession Header


wangxulong:
"authorizer": {
        "handler": "remote_json",
        "config": {
          "remote": "http://192.168.124.18:4000/authz/permission/allow",
          "payload": "{\"subject\": \"{{ print .Subject }}\", \"url\": \"{{ print .MatchContext.URL}}\",\"clientId\": \"{{ print .Extra.client_id }}\",\"method\": \"{{ print .Header }}\" }"
        }
      },

payload : subject ,url, clientId，is ok,  **but the method value from Header is empty** .

aeneasr:
Thank you for contributing to this repository by creating an issue!

Unfortunately, your issue lacks vital information, such as log files, the error message, the software version, your configuration or other pieces of the puzzle.

Please also ensure that your issue is appropriately formatted. If you do not know how to write markdown, you can find help [here](https://guides.github.com/features/mastering-markdown/).

Helping you with your problem is only possible if you share this information, and it will save a lot of time of back and forth on your as well as our end!

For this reason, this repository uses [issue templates](../tree/master/.github/ISSUE_TEMPLATE) which you can select when pressing "New issue". Please use one of those issue templates to fill in the required information. You can either create a new issue for this purpose and close this one, or leave a comment. 

> Do not edit the original post as we will not be notified when you do so.

If you do not provide the requested information, this issue will be closed.

k9ert:
I can reproduce the issue:
```
Version:    v0.38.6-beta.1
Git Hash:   a0c4d7fc46151000b4ae9db5a958b94611c9cd58
Build Time: 2021-01-27T09:26:59Z
```

oathkeeper config like this:
```
serve:
  proxy:
    port: 4455 # run the proxy at port 4455
  api:
    port: 4456 # run the api at port 4456

access_rules:
  repositories:
    - file://./rules_authorizer-remote-json.yaml

errors:
  fallback:
    - json
  handlers:
    json:
      enabled: true
      config:
        verbose: true
    redirect:
      enabled: true
      config:
        to: https://www.ory.sh/docs

mutators:
  noop:
    enabled: true

authorizers:
  remote_json:
    enabled: true
    config:
      remote: http://localhost:5000/oathkeeper_authorizer/remote_json
      payload: |
        {
          "subject": "{{ print .Subject }}", 
          "url": "{{ print .MatchContext.URL}}",
[...]

PR #718, opened 2021-05-05
fix: Initialize Header for AuthenticationSession

msumit:
## Related issue

https://github.com/ory/oathkeeper/issues/512

## Proposed changes

After wasting a lot of time on why I'm not able to use Headers within the Remote Authorizer, I found out that Oathkeeper isn't setting the Headers while initializing the AuthenticationSession. We use Remote Authorizer and pass on some key information through headers only & without this fix we can't use Oathkeeper at all. 

## Checklist

      the changes.
      works.

### DIFF
--- a/proxy/request_handler.go
+++ b/proxy/request_handler.go
@@ -337,6 +337,7 @@ func (d *RequestHandler) InitializeAuthnSession(r *http.Request, rl *rule.Rule) 
 
 	session := &authn.AuthenticationSession{
 		Subject: "",
+		Header: r.Header,
 	}
 
 	values, err := rl.ExtractRegexGroups(d.c.AccessRuleMatchingStrategy(), r.URL)
--- a/proxy/request_handler_test.go
+++ b/proxy/request_handler_test.go
@@ -44,7 +44,12 @@ import (
 )
 
 func newTestRequest(u string) *http.Request {
-	return &http.Request{URL: x.ParseURLOrPanic(u), Method: "GET"}
+	req, err := http.NewRequest("GET", u, nil)
+	if err != nil {
+		//do something
+	}
+	req.Header.Add("Test-Header", "Test-Value")
+	return req
 }
 
 func TestHandleError(t *testing.T) {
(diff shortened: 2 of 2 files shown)


aeneasr (changes requested):
This is a misunderstanding of the purpose of this field. The header field is used *in the response*. Setting the request headers as the default will mean that the server replies with the HTTP request headers in the response, which results in an invalid response.

I assume this was explained incorrectly in some documentation? If so, where was it so that we can fix it right away?

Regarding the feature you're looking for, I suggest to move the *request* header to `MatchContext`. We should also rename `Header` to `ResponseHeader` to avoid confusion. However, this is a breaking change for existing templates and warrants a new minor release if done.

aeneasr:
Oh and thank you for the contrib and sorry for the frustrating experience :)

aeneasr:
> This is a misunderstanding of the purpose of this field. The header field is used _in the response_. Setting the request headers as the default will mean that the server replies with the HTTP request headers in the response, which results in an invalid response.
> 
> I assume this was explained incorrectly in some documentation? If so, where was it so that we can fix it right away?
> 
> Regarding the feature you're looking for, I suggest to move the _request_ header to `MatchContext`. We should also rename `Header` to `ResponseHeader` to avoid confusion. However, this is a breaking change for existing templates and warrants a new minor release if done.

msumit:
Thanks, @aeneasr for a quick review. So if you look at the documentation of [remote authorizer](https://www.ory.sh/oathkeeper/docs/pipeline/authz#remote), it implies that you can use all fields of `AuthenticationSession` to customize the headers. Also, I'm not sure how we can use `MatchContext` to solve our use case? We need to forward `Cookie` header to the remote authorizer like this `"Cookie": "{{ .Header.Get \"Cookie\" }}"`. So another fix would be to auto-add `Cookie` header just like it's adding `Authorization` header?

aeneasr:
Yeah that's kinda confusing, the content is correct but it can lead one to believe that `Headers` contains the original HTTP request headers.

What I meant with using `MatchContext` was this:

```
"Cookie": "{{ .MatchContext.Header.Get \"Cookie\" }}"
```

So basically adding a new field what does what you want!

msumit:
Thanks @aeneasr, I thought `MatchContext` contains only a matching pattern & url, but if it can be used for Headers as well, then we may not need this change.

aeneasr:
Exactly, my point was adding a new field `Request` to `MatchContext` that does what you want

Issue #712, opened 2021-04-30
OAuth2 Introspection Authenticator requests new pre auth token for every request


pike1212:
**Describe the bug**

OAuth2 Introspection Authenticator requests new pre auth token for every request resulting in lots of token generation in the oauth server

**Expected behavior**

Resuse tokens until they expire

**Additional context**

I think fix is simple, don't rebuild roundTripper everytime config is called in AuthenticatorOAuth2Introspection

PR #723, opened 2021-05-10
fix: intro preauth client once

pike1212:
## Related issue

#712 

## Proposed changes

Only initializes oauth client once during config stage per config and doesn't initialize when creating new introspector.

## Checklist

      the changes.
      works.

### DIFF
--- a/pipeline/authn/authenticator_oauth2_introspection.go
+++ b/pipeline/authn/authenticator_oauth2_introspection.go
@@ -262,46 +262,47 @@ func (a *AuthenticatorOAuth2Introspection) Config(config json.RawMessage) (*Auth
 		return nil, NewErrAuthenticatorMisconfigured(a, err)
 	}
 
-	var rt http.RoundTripper
+	if a.client == nil {
+		var rt http.RoundTripper
+		if c.PreAuth != nil && c.PreAuth.Enabled {
+			var ep url.Values
 
-	if c.PreAuth != nil && c.PreAuth.Enabled {
-		var ep url.Values
+			if c.PreAuth.Audience != "" {
+				ep = url.Values{"audience": {c.PreAuth.Audience}}
+			}
 
-		if c.PreAuth.Audience != "" {
-			ep = url.Values{"audience": {c.PreAuth.Audience}}
+			rt = (&clientcredentials.Config{
+				ClientID:       c.PreAuth.ClientID,
+				ClientSecret:   c.PreAuth.ClientSecret,
+				Scopes:         c.PreAuth.Scope,
+				EndpointParams: ep,
+				TokenURL:       c.PreAuth.TokenURL,
+			}).Client(context.Background()).Transport
 		}
 
-		rt = (&clientcredentials.Config{
-			ClientID:       c.PreAuth.ClientID,
-			ClientSecret:   c.PreAuth.ClientSecret,
-			Scopes:         c.PreAuth.Scope,
-			EndpointParams: ep,
-			TokenURL:       c.PreAuth.TokenURL,
-		}).Client(context.Background()).Transport
-	}
-
-	if c.Retry == nil {
-		c.Retry = &AuthenticatorOAuth2IntrospectionRetryConfiguration{Timeout: "500ms", MaxWait: "1s"}
-	} else {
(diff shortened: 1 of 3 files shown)


pipeline/authn/authenticator_oauth2_introspection.go
```diff
@@ -252,61 +252,66 @@ func (a *AuthenticatorOAuth2Introspection) Validate(config json.RawMessage) erro
...
+func (a *AuthenticatorOAuth2Introspection) Config(config json.RawMessage) (*AuthenticatorOAuth2IntrospectionConfiguration, *http.Client, error) {
 	var c AuthenticatorOAuth2IntrospectionConfiguration
 	if err := a.c.AuthenticatorConfig(a.GetID(), config, &c); err != nil {
-		return nil, NewErrAuthenticatorMisconfigured(a, err)
+		return nil, nil, NewErrAuthenticatorMisconfigured(a, err)
 	}
 
-	var rt http.RoundTripper
+	clientKey := fmt.Sprintf("%x", md5.Sum([]byte(config)))
+	client, ok := a.clientMap[clientKey]
```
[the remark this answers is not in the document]
pike1212: TIL, let me know how that looks

pipeline/authn/authenticator_oauth2_introspection.go
```diff
@@ -252,61 +254,71 @@ func (a *AuthenticatorOAuth2Introspection) Validate(config json.RawMessage) erro
...
 
-	var rt http.RoundTripper
+	clientKey := fmt.Sprintf("%x", md5.Sum([]byte(config)))
+	a.mu.RLock()
+	client, ok := a.clientMap[clientKey]
+	a.mu.RUnlock()
 
-	if c.PreAuth != nil && c.PreAuth.Enabled {
-		var ep url.Values
+	if !ok {
```
aeneasr, line 272: Should we add a test for this? I imagine a test which creates one instance for `AuthenticatorOAuth2Introspection`, uses three configs (one with preauth, one without preauth, one with another preauth) and then calls `Config` and expecnts the first config to always produce the same client (checking the pointer), the second never returns a client, and the third returns a different client. To make order random I would probably call config 1, 2, 3 - and then 2, 3, 1 to ensure this does not have side effects.

pike1212: I added a test, let me know what you think...

pipeline/authn/authenticator_oauth2_introspection_test.go
```diff
@@ -589,4 +590,61 @@ func TestAuthenticatorOAuth2Introspection(t *testing.T) {
...
+		if err != nil {
+			require.NoError(t, err)
+		}
+
+		_, preauthTwoClient, err := authenticator.Config(preAuthConfigTwo)
+		if err != nil {
+			require.NoError(t, err)
+		}
+
+		if noPreauthClient == preauthOneClient || noPreauthClient == preauthTwoClient || preauthOneClient == preauthTwoClient {
```
aeneasr, line 617: You could simplify this with:

```suggestion
		require.NotEqual(t, noPreauthClient, preauthOneClient )
		// ...
```

pipeline/authn/authenticator_oauth2_introspection_test.go
```diff
@@ -589,4 +590,61 @@ func TestAuthenticatorOAuth2Introspection(t *testing.T) {
...
+		if err != nil {
+			require.NoError(t, err)
+		}
+		if noPreauthClient2 != noPreauthClient {
+			t.FailNow()
+		}
+
+		if noPreauthClient == preauthOneClient || noPreauthClient == preauthTwoClient || preauthOneClient == preauthTwoClient {
+			t.FailNow()
+		}
```
aeneasr, line 647: Same here :)

aeneasr (commented):
Awesome! I think we can use require/assert for the equals comparison but otherwise this is exactly what I imagined! thank you :)

pike1212:
@aeneasr can this be merged?

aeneasr:
Yes! :)

aeneasr:
Thank you! :)

Issue #203, opened 2019-05-29
Decouple credentials fetch from actual requests


stszap:
**Is your feature request related to a problem? Please describe.**

First of all thanks for the big update in 0.16, it made our setup much simpler. Our issue is that timeout for credentials fetcher is now fixed (we use [auth0](https://auth0.com) with jwt authenticator). Unfortunately, requests to jwks.json are very slow (1-1.5s from countries where our developers reside) and occasionally we get 403 from oathkeeper, but the next request succeeds.

The actual scenario is (I believe):

1. oathkeeper starts, empty caches, no connections
1. client makes a request to an endpoint with jwt authenticator
1. oathkeeper makes a request to jwks url (synchronously with the client's request)
1. 1s passes and ctx timeout occurs, the client gets`{"error":{"code":403,"status":"Forbidden","reason":"An internal server error occurred, please contact the system administrator","message":"Access credentials are not sufficient to access this resource"}}`
1. oathkeeper actually finishes the request, caches the key set and even keeps tcp connection to jwks server open
1. the client makes another request
1. oathkeeper gets the key from cache and proxies the request without an error
1. 30s passes and the client makes another request
1. the cache for jwks has expired and oathkeeper make another request but this time the connection is still open so there is no tcp handshake and request takes less than 1s
1. client's request gets proxied without an error but with some added latency
1. if there are no requests for a while oathkeeper closes tcp connection and the whole process repeats

**Describe the solution you'd like**

An ideal solution would be some kind of background function that will fetch jwks periodically so actual requests from clients would always use cache (even if it's stale). The benefits would be:

1. no added latency because no additional synchronous requests are made
1. higher error tolerance because even if jwks server is unavailable the cache is always used and when the server is up again the cache will be updated automatically

**Describe alternatives you've considered**

I understand that the ideal solution might be difficult to implement so I see 2 alternative ones:

- make the timeout configurable
- use something outside of oathkeeper (eg setup some external process that will save jwks to file system and use "file://" in oathkeeper)

**Additional context**

If I understand correctly the 1s limit is set here https://github.com/ory/oathkeeper/blob/1e03ee24b466b01fe4f81e5a4fda8ab42f9a0615/driver/registry_memory.go#L183 and then used here https://github.com/ory/oathkeeper/blob/1e03ee24b466b01fe4f81e5a4fda8ab42f9a0615/credentials/fetcher_default.go#L94

aeneasr:
[...]

PR #726, opened 2021-05-15
feat: improved JWT Authorizer JWKs fetching

eroznik:
## Related issue 

#203 
@aeneasr

## Proposed changes

The current implementation is not suitable for deployments in which low response times are required.  With this PR I am improving the current implementation with the following changes:
- When the `FetcherDefault` fetches new JWKS and the request takes more then allowed, it will use "stale" JWKs for JWT validation. The refresh request will still continue and if successful, it will refresh the cached JWK for the "next" request.
- `FetcherDefault` already accepts configuration for JWK cache TTL and max await for the refresh(and init!) JWK requests, the problem was that the end user could not change these values. This PR enables the user to change these values with 2 new configs: `authenticators.jwt.config.jwks_max_wait` and `authenticators.jwt.config.jwks_ttl`.

## Checklist

      the changes.
      works.

## Further comments

I am aware that adding 2 more configuration keys makes stuff a little bit more complex, but I am certain this will help out quite a bit of users down the road. Schemas + docs + reference config updated.

### DIFF
--- a/.schema/config.schema.json
+++ b/.schema/config.schema.json
@@ -574,7 +574,7 @@
             ]
           ]
         },
-        "jwk_max_wait": {
+        "jwks_max_wait": {
           "title": "Max await interval for the JWK fetch",
           "type": "string",
           "description": "The configuration which sets the max wait threshold when fetching new JWKs",
--- a/.schema/config.schema.json
+++ b/.schema/config.schema.json
@@ -574,6 +574,26 @@
             ]
           ]
         },
+        "jwk_max_wait": {
+          "title": "Max await interval for the JWK fetch",
+          "type": "string",
+          "description": "The configuration which sets the max wait threshold when fetching new JWKs",
+          "default" : "1s",
+          "examples": [
+            "100ms",
+            "1s"
+          ]
+        },
+        "jwk_ttl": {
+          "title": "JWK cache TTL configuration",
+          "type": "string",
+          "description": "The time interval for which fetched JWKs are cached",
+          "default" : "30s",
+          "examples": [
+            "30m",
+            "6h"
+          ]
+        },
         "scope_strategy": {
           "$ref": "#/definitions/scopeStrategy"
         },
--- a/.schema/config.schema.json
+++ b/.schema/config.schema.json
@@ -584,7 +584,7 @@
             "1s"
           ]
         },
-        "jwk_ttl": {
+        "jwks_ttl": {
           "title": "JWK cache TTL configuration",
           "type": "string",
           "description": "The time interval for which fetched JWKs are cached",
--- a/credentials/fetcher_default.go
+++ b/credentials/fetcher_default.go
@@ -52,6 +52,10 @@ type reasoner interface {
 	Reason() string
 }
 
+type fetchResult struct {
+	finishedWithTimeout bool
+}
+
 var _ Fetcher = new(FetcherDefault)
 
 type FetcherDefault struct {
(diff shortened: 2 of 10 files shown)


.schema/config.schema.json
```diff
@@ -574,6 +574,26 @@
             ]
           ]
         },
+        "jwk_max_wait": {
```
aeneasr, line 577: To keep consistency maybe do `jwks_`? :)

credentials/fetcher_default.go
```diff
@@ -124,19 +128,21 @@ func (s *FetcherDefault) fetchParallel(ctx context.Context, locations []url.URL)
 	select {
 	case <-ctx.Done():
 		s.l.Errorf("Ignoring JSON Web Keys from at least one URI because the request timed out waiting for a response.")
+		return fetchResult{true}
```
aeneasr, line 131: Maybe return an error here instead? And then we check for errors in the caller code? That would be a bit more idiomatic go code :)

credentials/fetcher_default_test.go
```diff
@@ -153,6 +153,14 @@ func TestFetcherDefault(t *testing.T) {
 		assert.True(t, check("8e884167-1300-4f58-8cc1-81af68f878a8"))
 	})
 
+	time.Sleep(maxWait * 7) // wait so the fetched key reaches ttl
+
+	t.Run("name=should find the previously fetched key if the refresh request times out", func(t *testing.T) {
+		key, err := s.ResolveKey(context.Background(), uris, "c61308cc-faef-4b98-99c3-839f513ac296", "sig")
```
aeneasr, line 159: How do we know if the key returned here is actually from the cache and not from the HTTP request which fetched it again?

aeneasr (commented):
Thank you! Just some minor details here :)

Also, sorry for taking so long to review. I just had time to work on the backlog...down to only 12 more GH notifications (from 500 😰 )

.schema/config.schema.json
eroznik: makes sense, changed to `jwks_`

credentials/fetcher_default.go
eroznik: Thanks for the honest feedback, I really appreciate comments like this - I am not that proficient with Go yet :)
Changed to `return error`, is now ok - or should I even improve the implementation with a custom error?

credentials/fetcher_default_test.go
eroznik: Nice catch, the test was a little bit shady - I improved it.

eroznik:
> Thank you! Just some minor details here :)
> 
> Also, sorry for taking so long to review. I just had time to work on the backlog...down to only 12 more GH notifications (from 500 cold_sweat )

No worries, I understood already from my first Oathkeeper PR that you're quite a busy man :))

credentials/fetcher_default.go
```diff
@@ -124,19 +124,21 @@ func (s *FetcherDefault) fetchParallel(ctx context.Context, locations []url.URL)
 	select {
 	case <-ctx.Done():
 		s.l.Errorf("Ignoring JSON Web Keys from at least one URI because the request timed out waiting for a response.")
+		return errors.Errorf("At least one JWT fetch took more then the allowed duration")
```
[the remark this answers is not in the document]
eroznik: Checked the docs and passed a few time with debugger, at the end i used `context.DeadlineExceeded` since that gets returned when the timeout is exceeded.

aeneasr (approved):
Awesome, thank you for your contribution! :))

PR #744, opened 2021-06-08
feat: Extended Root CA for client TLS connections (#181)

christian-roggia:
See https://github.com/ory/oathkeeper/pull/706.

### DIFF
--- a/docs/docs/api-access-rules.md
+++ b/docs/docs/api-access-rules.md
@@ -236,20 +236,33 @@ authenticators:
 
 ## Client TLS Transport
 
-In case you want to enable TLS connections to services protected by a self-signed certificate you will very likely need to setup your own custom certificates. Through the `client_tls` configuration you can set custom certificates that will be used by oathkeeper to connect to other services such as the upstream service, authorizers, authenticators, and mutators. The certificates are cached in memory to reduce latency, you can set the cache expiration conditions through the `cache` configuration or disable caching entirely. NOTE: Caching is enabled by default.
+In case you want to enable TLS connections to services protected by a
+self-signed certificate you will very likely need to setup your own custom
+certificates. Through the `client_tls` configuration you can set custom
+certificates that will be used by oathkeeper to connect to other services such
+as the upstream service, authorizers, authenticators, and mutators. The
+certificates are cached in memory to reduce latency, you can set the cache
+expiration conditions through the `cache` configuration or disable caching
+entirely. NOTE: Caching is enabled by default.
 
-Use the `cache.refresh_frequency` to test once every `1000` requests for a certificate file size or modification timestamp change. Setting the value to 0 will disable the refresh on a frequency functionality, but retrain the update on file path change functionality.
+Use the `cache.refresh_frequency` to test once every `1000` requests for a
+certificate file size or modification timestamp change. Setting the value to 0
+will disable the refresh on a frequency functionality, but retrain the update on
+file path change functionality.
 
-Use the `cache.ttl` to force the refresh of certificates once every `5m` regardless of the cache status. Setting this value to 0 will disable forced cache eviction due to an expired time-to-live.
+Use the `cache.ttl` to force the refresh of certificates once every `5m`
+regardless of the cache status. Setting this value to 0 will disable forced
+cache eviction due to an expired time-to-live.
 
 **oathkeeper.yml**
+
 ```yaml
 serve:
   proxy:
     client_tls:
       trusted_certificates:
-        - "/my-certs/certificate-to-append.crt"
-        - "/my-certs/another-certificate.crt"
+        - /my-certs/certificate-to-append.crt
+        - /my-certs/another-certificate.crt
       cache:
         refresh_frequency: 1000
         ttl: 5m
--- a/docs/docs/api-access-rules.md
+++ b/docs/docs/api-access-rules.md
@@ -234,19 +234,25 @@ authenticators:
 }
 ```
 
-## Upstream Configuration
+## Upstream Transport Configuration
 
-Upstreams are often protected by SSL certificates trusted by the SSL root servers, but what if you want to run your upstream infrastructure using self-signed certificates? Use the `ca_append_crt_path` to specifiy the certificate file to append to the Root Certificate Authority (CA) of the upstream transport. Upstream transport is cached to reduce IO overhead at scale. The upstream transport certificate will automatically update if the certificate file path is changed.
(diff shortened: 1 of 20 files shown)


docs/docs/api-access-rules.md
```diff
@@ -234,6 +234,27 @@ authenticators:
 }
 ```
 
+## Upstream Transport Configuration
+
```
aeneasr, line 238: Probably makes sense here to document that this is supported both for proxy upstreams as well as for all authenticators, authorizers, ... that do HTTP?

internal/certs/cert_manager.go
```diff
@@ -0,0 +1,177 @@
...
+func NewCertManager(c configuration.Provider) *CertManager {
+	return &CertManager{
+		c:         c,
+		isWindows: runtime.GOOS == "windows",
+	}
+}
+
+func (cm *CertManager) systemCertPool() (*x509.CertPool, error) {
+	if cm.isWindows {
+		return x509.NewCertPool(), nil
```
aeneasr, line 54: Does windows still use the system cert pool though?

aeneasr:
@christian-roggia marking this as a draft because you said it's WIP on slack :)

internal/certs/cert_manager.go
christian-roggia: Windows does not support the system certificate pool. Calling `SystemCertPool` will always return the error `errors.New("crypto/x509: system root pool is not available on Windows")`

https://golang.org/src/crypto/x509/cert_pool.go?s=3079:3119#L95

christian-roggia:
@aeneasr this PR should be cleaned up and ready for review. I deployed this version to our cluster and works like a charm with self-signed TLS certificates. As soon as this is merged I will open a PR also for the Helm charts.

aeneasr:
Thank you, this looks great! The CI is failing because some files are formatted incorrectly. To format them, run:

```
$ make format
$ git commit -a -m "styles: format code"
$ git push
```

Thank you!

christian-roggia:
As I wrote on Slack, I run the command before but I didn't commit the two CHANGELOG files as I didn't touch them and the number of changed lines was huge. Looks like it worked tough now that I committed also those two files, I guess the linter was recently updated?

Thank you anyway for the help!

aeneasr:
The problem was that I changed something in the CI which pushed an unformatted commit to master. Your PR includes those changes thus had issues in the linter. The issue is resolved in the CI but since then no update was triggered on Ory Oathkeeper.

I'll be reviewing your PR shortly!

internal/certs/cert_manager.go
```diff
@@ -0,0 +1,183 @@
...
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/pkg/errors"
+
+	"github.com/ory/oathkeeper/driver/configuration"
+)
+
+type CertManager struct {
```
aeneasr, line 17: Could you please add some tests to the certmanager? Also, what should we take care of here. For example, test different cert pools? How would this look like on Windows? :)

I haven't worked with x509 cert pools before. We should also generally ensure that:

1. The cache is used
2. The cache is invalidated properly
3. No untrusted certs are being used
4. The OS cert store is being used appropriately on the different OS'es (could be difficult for mac as we don't have a pipeline for that...)
5.

docs/docs/api-access-rules.md
```diff
@@ -234,6 +234,40 @@ authenticators:
...
+regardless of the cache status. Setting this value to 0 will disable forced
+cache eviction due to an expired time-to-live.
+
+**oathkeeper.yml**
+
+```yaml
+serve:
+  proxy:
+    client_tls:
+      trusted_certificates:
```
aeneasr, line 263: If this is used for other upstream services (authenticators, ...) I think it would make sense to have `client_tls` as a root level config item

driver/configuration/provider.go
```diff
@@ -40,6 +41,9 @@ type Provider interface {
 	ProxyReadTimeout() time.Duration
 	ProxyWriteTimeout() time.Duration
 	ProxyIdleTimeout() time.Duration
+	ProxyServeClientTLSTrustedCerts() []string
+	ProxyServeClientTLSCacheRefreshFrequency() int
+	ProxyServeClientTLSCacheTimeToLive() time.Duration
```
aeneasr, line 46: If we move the config to the root level we need to change this

driver/configuration/provider_viper.go
```diff
@@ -43,6 +43,9 @@ const (
 	ViperKeyProxyIdleTimeout                    = "serve.proxy.timeout.idle"
 	ViperKeyProxyServeAddressHost               = "serve.proxy.host"
 	ViperKeyProxyServeAddressPort               = "serve.proxy.port"
+	ViperKeyProxyClientTLSTrustedCerts          = "serve.proxy.client_tls.trusted_certificates"
+	ViperKeyProxyClientTLSCacheRefreshFrequency = "serve.proxy.client_tls.cache.refresh_frequency"
+	ViperKeyProxyClientTLSCacheTimeToLive       = "serve.proxy.client_tls.cache.ttl"
```
aeneasr, line 48: If we move the config to the root level we need to change this

internal/certs/round_tripper.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+var _ http.RoundTripper = (*RoundTripper)(nil)
+
+type RoundTripper struct {
+	cm *CertManager
+	tr *http.Transport
+}
+
+func NewRoundTripper(cm *CertManager) *RoundTripper {
```
aeneasr, line 15: Some tests would be great here as well

pipeline/authn/authenticator_bearer_token.go
```diff
@@ -8,9 +8,11 @@ import (
 	"github.com/tidwall/gjson"
```
aeneasr, line 8: Could we ensure in all the authenticators that the new TLS config is actually being used by adding tests for that case?

proxy/proxy_test.go
```diff
@@ -429,6 +431,61 @@ func TestConfigureBackendURL(t *testing.T) {
 	}
 }
 
+func TestUpstreamAppendCaCertToRootCa(t *testing.T) {
```
aeneasr, line 434: I think we could make this a generic test helper which could then be used in the different relevant places (authenticators, mutators, proxy, authorizers)

aeneasr (commented):
Thank you for advancing this! From the documentation and overall implementation this looks very good!

To confirm that everything works as expected I think we definitely need to add some more tests.

aeneasr:
Just checking in if you need help with the tests :)

christian-roggia:
I will be updating the configuration path to be at the root level but I will most definitely need help setting up tests!

aeneasr:
Ok! Ping me when you're ready to work on it :) Meanwhile I will mark it as a draft. That declutters our review backlog :)

Once you're done with your changes and would like someone to review them, mark the PR as ready and request a review from one of the maintainers.

Thank you!

aeneasr:
As there has not been an update in the past 6 month, we'll consider this one stale and I'll close the PR. But please, do not feel discouraged in continuing the work if you have some time, the change is still appreciated! :)

PR #750, opened 2021-06-17
chore: update README

vinckr:
Changed the README to be identical to the one used in hydra,kratos ect.


aeneasr:
Add best practices here as well?

Issue #565, opened 2020-11-04
Hydator Mutator Client Credential


schreddies:
**Is your feature request related to a problem? Please describe.**

The hydrator have only Basic auth - see config: https://www.ory.sh/oathkeeper/docs/pipeline/mutator/#configuration-4, however for more flexibility, similiar `pre-authorization` mechanism could ba added as in the instropsection authenticator handler: https://www.ory.sh/oathkeeper/docs/pipeline/mutator/#configuration-4

**Describe the solution you'd like**

Adding the whole `pre-authorization` config to the `hydrator` will increase possibilities to use hydrator in multiple cases.

**Describe alternatives you've considered**

Basic auth cannot be used with current config for client credentials for external hydrator service.

**Additional context**

Can contribute to that.

aeneasr:
Hey there, I think this is a good idea! We have a call this week anyways, correct? Let's discuss the ideas there!

aeneasr:
Incorrect stalebot detection - this was assigned a milestone.

PR #751, opened 2021-06-17
Oauth2 client credential flow support for hydrator mutator

schreddies:
## Related issue

https://github.com/ory/oathkeeper/issues/565

## Proposed changes

Using the Oauth2 Client Credential Flow will beneficial to operate and include other services, protected with oauth2. for some users to hydrate (enrich) the pipeline flow in the Oathkeeper.

## Checklist

- [X ] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md).
- [X ] I have read the [security policy](../security/policy).
- [X ] I confirm that this pull request does not address a security
      the changes.
      works.

## Further comments

No tests because of the nature of this change - it seems that I am not able to write tests without including 3rd party to issue the access tokens. I may require some help with documentation especially as I am not native speaker.

### DIFF
--- a/pipeline/mutate/mutator_hydrator.go
+++ b/pipeline/mutate/mutator_hydrator.go
@@ -22,6 +22,7 @@ package mutate
 
 import (
 	"bytes"
+	"context"
 	"crypto/md5"
 	"encoding/json"
 	"fmt"
@@ -30,6 +31,7 @@ import (
 	"time"
 
 	"github.com/dgraph-io/ristretto"
+	"golang.org/x/oauth2/clientcredentials"
 
 	"github.com/ory/oathkeeper/pipeline/authn"
 	"github.com/ory/oathkeeper/x"
@@ -67,6 +69,13 @@ type BasicAuth struct {
 	Password string `json:"password"`
 }
 
+type MutatorHydratorRetryConfiguration struct {
+	Timeout string `json:"max_delay"`
+	MaxWait string `json:"give_up_after"`
+}
+
+//new
+
 type auth struct {
 	Basic BasicAuth `json:"basic"`
 }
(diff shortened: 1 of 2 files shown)


pipeline/mutate/mutator_hydrator.go
```diff
@@ -254,6 +275,40 @@ func (a *MutatorHydrator) Config(config json.RawMessage) (*MutatorHydratorConfig
 		return nil, NewErrMutatorMisconfigured(a, err)
 	}
 
+	var rt http.RoundTripper
+
+	if c.PreAuth != nil && c.PreAuth.Enabled {
```
aeneasr, line 280: Is this copy/pasted from another handler? If so, I think it would make sense to extract that logic into a common module which then is imported and used here :)

aeneasr (commented):
Thank you for the contribution! I think this is going into the right direction.

It would be awesome if you could add tests! :relieved:

Tests are important to open source projects because the allow reviewers to better judge if the changes are doing what was intended. They also ensure that your changes never regress / break when someone else works on code that is related to this.

If you need help with the tests feel free to push your work in progress and ping myself or another maintainer :raising_hand:

aeneasr:
While the PR is being worked on I will mark it as a draft. That declutters our review backlog :)

Once you're done with your changes and would like someone to review them, mark the PR as ready and request a review from one of the maintainers.

Thank you!

pipeline/mutate/mutator_hydrator.go
schreddies: It highly influenced by the client credential from `oauth2_interceptor` configuration `pre_auth - @aeneasr you are totally right! It makes more sense.

schreddies:
@aeneasr Thank you for review! I will do changes and mark it `Ready to review`. Thank you @aeneasr !

Issue #491, opened 2020-07-30
oauth2_introspection not parsing single string aud valie


stevepnbscott:
**Describe the bug**

I have configured oathkeeper to use an authenticator for oauth2_introspection using keycloak.

By default keycloak returns only a single value in the aud claim of the token;

```
{
    "exp": 1596123440,
    "nbf": 0,
    "iat": 1596123140,
    "jti": "57f536d3-2b17-4652-9859-c58a8f4c77a7",
    "aud": "account",
    "typ": "Bearer",
    "acr": "1",
    "active": true
}
```

According to the RFC  this is valid : 

> _aud
>
>      OPTIONAL.  Service-specific string identifier or list of string
>      identifiers representing the intended audience for this token, as
>      defined in JWT [RFC7519]._

However I see Oathkeeper unable to parse that return value; 

```
**_time=2020-07-30T14:27:31Z level=error msg=An error occurred while handling a request code=500 debug= details=map[] error=json: cannot unmarshal string into Go struct field AuthenticatorOAuth2IntrospectionResult.aud of type []string reason= request-id= status=500 writer=JSON_*
```

**Reproducing the bug**

using simple docker file : 

```
FROM oryd/oathkeeper:latest1

ADD config.yaml /config.yaml
ADD rules.json /rules.json
ADD jwks.json /jwks.json
```

1. Run oathkeeper locally;

```
$ docker build -t ory-oathkeeper-demo .

# Docker run with mapping to config with rules/jwsk etc locally
docker run --rm \
  --name ory-oathkeeper-demo \
  -v ~/Oathkeeper:/configs \
  -p 4455:4455 \
  -p 4456:4456 \
  ory-oathkeeper-demo \
  --config /configs/config.yaml \
  serve
```

2. Call the API to trigger the introspeciton

```
$ curl -X GET -H 'Authorization: Bearer <token here>' http://127.0.0.1:4455/auth/intro
```

_Server logs_

```
time=2020-07-30T14:27:31Z level=info msg=started handling request http_request=map[headers:map[accept:*/* authorization:Value is sensitive and has been redacted. To see the value set config key "log.leak_sensitive_values = true" or environment variable "LOG_LEAK_SENSITIVE_VALUES=true". user-agent:curl/7.54.0] host:127.0.0.1:4455 method:GET path:/auth/intro query:<nil> remote:172.17.0.1:56148 scheme:http]
[...]

PR #792, opened 2021-07-17
fix: introspection response

tjandrayana:
I try to fixing the issue related oauth2_introspection not parsing single string aud value

Based on this protocol https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation the audience field need to changed, to receive []string or string. 

```REQUIRED. Audience(s) that this ID Token is intended for. It MUST contain the OAuth 2.0 client_id of the Relying Party as an audience value. It MAY also contain identifiers for other audiences. In the general case, the aud value is an array of case sensitive strings. In the common special case when there is one audience, the aud value MAY be a single case sensitive string.
```
There are some solutions we can implement : 

1. Change the audience as an interface 
  -  https://github.com/ory/oathkeeper/blob/e4e2263c7b97e47506919e29442efb395eaba99d/pipeline/authn/authenticator_oauth2_introspection.go#L90
  Effort : 
  - Need to create function to check the type of interface, string or []string
  - Need to update the current unit testing, possible break the current unit test
  
2. Create new struct and then set the audience as a single string, so if there is possibility failed to unmarshal / casting and then unmarshal / cast with this new struct
  Effort :
 - Create new function to modify the response
 - Not break the current unit testing

So, because of that, I decide to create new struct named Fork and add the function modify struct in this PR.
With this Fork struct and this function the unit testing also work fine after I execute in my local.
Hopefully with this PR can help everyone that face same issue.
Thank you

## Related issue(s)

https://github.com/ory/oathkeeper/issues/491

## Checklist

- [x] I am following the
      the changes.
      works.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ node_modules/
 LICENSE.txt
 *-packr.go
 dev
+local-conf/
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,3 @@ node_modules/
 LICENSE.txt
 *-packr.go
 dev
-local-conf/
--- a/pipeline/authn/authenticator_oauth2_introspection.go
+++ b/pipeline/authn/authenticator_oauth2_introspection.go
@@ -96,7 +96,7 @@ type AuthenticatorOAuth2IntrospectionResult struct {
 	TokenUse  string                 `json:"token_use"`
 }
 
-// based on
+// Audience value mau be a single sensitive string.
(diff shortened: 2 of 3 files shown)


aeneasr (changes requested):
Thank you for the PR! However, unfortunately, I do not believe that it will resolve what you are trying to do, because the type assertion will always be false. It might make sense to use something like `gjson.Get` to see if you have a string or an array in the audience. It would also make sense to add an e2e test to ensure that the feature works as expected :)

PR #811, opened 2021-08-24
chore: store oauth2 introspection result as bytes in cache

j3r0lin:
OAuth2 introspection result are using raw value to be stored in cache, but the cached value may be edit after stored into cache, which may cause go map concurrent write issue, so I changed the cached value to bytes using json serializer. 

## Checklist

- [ ] I am following the
      the changes.
      works.

### DIFF
--- a/pipeline/authn/authenticator_oauth2_introspection_cache_test.go
+++ b/pipeline/authn/authenticator_oauth2_introspection_cache_test.go
@@ -1,0 +1,71 @@
+package authn
+
+import (
+	"github.com/ory/fosite"
+	"github.com/ory/oathkeeper/driver/configuration"
+	"github.com/ory/viper"
+	"github.com/ory/x/logrusx"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"testing"
+	"time"
+)
+
+func TestCache(t *testing.T) {
+	viper.Reset()
+
+	viper.Set("authenticators.oauth2_introspection.config.cache.enabled", true)
+	viper.Set("authenticators.oauth2_introspection.config.introspection_url", "http://localhost:8080/")
+
+	logger := logrusx.New("", "")
+	c := configuration.NewViperProvider(logger)
+	a := NewAuthenticatorOAuth2Introspection(c, logger)
+	assert.Equal(t, "oauth2_introspection", a.GetID())
+
+	config, _, err := a.Config(nil)
+	require.NoError(t, err)
+
+	t.Run("method=tokenToCache", func(t *testing.T) {
+		t.Run("case=cache value", func(t *testing.T) {
+			i := &AuthenticatorOAuth2IntrospectionResult{
+				Active: true,
+				Extra:  map[string]interface{}{"extra": "foo"},
+			}
+
+			a.tokenToCache(config, i, "token", fosite.WildcardScopeStrategy)
+			// wait cache to save value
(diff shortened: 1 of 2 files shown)


aeneasr (commented):
Thank you for the contribution! I think this is going into the right direction.

It would be awesome if you could add tests! :relieved:

Tests are important to open source projects because the allow reviewers to better judge if the changes are doing what was intended. They also ensure that your changes never regress / break when someone else works on code that is related to this.

j3r0lin:
The value to be stored is created from the introspection call result, which may always could be marshaled to bytes, I have no idea how to generate a scenario to mock the marshal error branch. Do you have any suggestion about how to write this test case?

aeneasr:
Sorry for the late reply - the best way would probably be to create a test which modifies the struct after it was cached and then the cached value is incorrect!

j3r0lin:
@aeneasr I think it's ready for review now.

aeneasr (approved):
Great job, thank you for your contribution!

pipeline/authn/authenticator_oauth2_introspection_cache_test.go
```diff
@@ -0,0 +1,91 @@
...
+	t.Run("method=tokenToCache", func(t *testing.T) {
+		t.Run("case=cache value", func(t *testing.T) {
+			i := &AuthenticatorOAuth2IntrospectionResult{
+				Active: true,
+				Extra:  map[string]interface{}{"extra": "foo"},
+			}
+
+			a.tokenToCache(config, i, "token", fosite.WildcardScopeStrategy)
+			// wait cache to save value
+			time.Sleep(time.Millisecond * 10)
```
aeneasr, line 38: If you update ristretto to the newest version it has a new method called wait which ensures that the cache worker has completed it’s work. It is more reliable and less flaky than waiting here! Could you maybe use that instead? :)

aeneasr (changes requested):
Sorry, I noticed one more thing!

pipeline/authn/authenticator_oauth2_introspection_cache_test.go
j3r0lin: update ristretto to v0.1.0 will cause an issue: flag redefined: v.
which defined in :

```go
# github.com/google/martian/init.go
var (
	level = flag.Int("v", 0, "log level")
)
```

```go
# github.com/google/golang/glog.go
func init() {
	flag.BoolVar(&logging.toStderr, "logtostderr", false, "log to standard error instead of files")
	flag.BoolVar(&logging.alsoToStderr, "alsologtostderr", false, "log to standard error as well as files")
	flag.Var(&logging.verbosity, "v", "log level for V logs")
	flag.Var(&logging.stderrThreshold, "stderrthreshold", "logs at or above this threshold go to stderr")
	flag.Var(&logging.vmodule, "vmodule", "comma-separated list of pattern=N settings for file-filtered logging")
	flag.Var(&logging.traceLocation, "log_backtrace_at", "when logging hits line file:N, emit a stack trace")

	// Default stderrThreshold is ERROR.
	logging.stderrThreshold = errorLog

	logging.setVState(0, nil, false)
	go logging.flushDaemon()
}
```

aeneasr: 🤦‍♂️ ok then keeping it as is is fine! Thank you :)

Issue #786, opened 2021-07-09
Authenticator bearer_token token_from query_parameter does not forward token to check_session_url if preserve_path: true


hunzikerj:
**Describe the bug**

The bearer_token authenticator does only forward request method/path/headers. If preserve_path is set to true then the query_parameter will not get forwarded as the url is replaced by the check_session_url.

**Reproducing the bug**

Using Oathkeeper according to the example in the documentation.

```
# Some Access Rule Preserving Path: access-rule-2.yaml
id: access-rule-2
# match: ...
# upstream: ...
authenticators:
  - handler: bearer_token
    config:
      check_session_url: https://session-store-host/check-session
      token_from:
        query_parameter: auth-token
      preserve_path: true
```

**Expected behavior**

If preserve_path is set to true then the query_parameter should be appended to the check_session_url in the request.

Example:

`https://session-store-host/check-session?auth-token=xxx`

**Environment**

- Version: v0.38.12-beta.1

aeneasr:
Yup, that makes sense - tracking as a feat request :)

grantzvolsky:
The `preserve_path` flag preserves the path in `check_session_url` because by default it is the path in the request that is forwarded. The counterintuitive property/bug is that while the request path is forwarded by default, the request query is ignored by default, without any option to forward it.

I will submit a PR that forwards the request query by default, which is consistent with forwarding the path by default. This is a breaking change because someone may depend on the query in `check_session_url` being preserved as it currently is. Should I introduce a new flag, `preserve_query`, and document it as a breaking change?

grantzvolsky:
We could also make the `preserve_query` flag true by default, which would achieve the same effect without any breaking changes.

PR #817, opened 2021-09-06
fix: allow forwarding query parameters to the session store

grantzvolsky:
This commit adds a new flag, preserve_query, in order to resolve issue #786. It
would be ideal for this flag to be false by default for consistency with
preserve_path, but it is true by default in order to preserve backwards
compatibility.

## Checklist

- [x] I am following the
      the changes.
      works.

### DIFF
--- a/pipeline/authn/authenticator_cookie_session_test.go
+++ b/pipeline/authn/authenticator_cookie_session_test.go
@@ -32,7 +32,7 @@ func TestAuthenticatorCookieSession(t *testing.T) {
 		t.Run("description=should fail because session store returned 400", func(t *testing.T) {
 			testServer, _ := makeServer(400, `{}`)
 			err := pipelineAuthenticator.Authenticate(
-				makeRequest("GET", "/", map[string]string{"sessionid": "zyx"}, ""),
+				makeRequest("GET", "/", "", map[string]string{"sessionid": "zyx"}, ""),
 				session,
 				json.RawMessage(fmt.Sprintf(`{"check_session_url": "%s"}`, testServer.URL)),
 				nil,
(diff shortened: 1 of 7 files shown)


aeneasr:
Let me know when this is ready for a pair of eyes :)

grantzvolsky:
This is ready to be reviewed now :)

pipeline/authn/authenticator_cookie_session_test.go
```diff
@@ -58,7 +58,7 @@ func TestAuthenticatorCookieSession(t *testing.T) {
 		t.Run("description=should pass through method, path, and headers to auth server", func(t *testing.T) {
 			testServer, requestRecorder := makeServer(200, `{"subject": "123"}`)
 			err := pipelineAuthenticator.Authenticate(
-				makeRequest("PUT", "/users/123?query=string", map[string]string{"sessionid": "zyx"}, ""),
+				makeRequest("PUT", "/users/123", "query=string", map[string]string{"sessionid": "zyx"}, ""),
 				session,
 				json.RawMessage(fmt.Sprintf(`{"check_session_url": "%s"}`, testServer.URL)),
```
aeneasr, line 63: Could you please add a test which uses a `check_session_url` that has a query parameter and observes the behaviour dependent on preserve_query?

aeneasr (changes requested):
Looking great! I think we should add test cases which ensures that query parameters from `check_session_url` are indeed preserved (not just empty but actually added) when `preserve_query` is set.

Otherwise, this looks great!

pipeline/authn/authenticator_cookie_session_test.go
grantzvolsky: True! It should be covered now :)

aeneasr (approved):
Awesome, thank you! 🎉

PR #601, opened 2020-12-03
WIP: fix: oauth2_introspection parsing single string aud value

daviddelucca:
## Related issue

https://github.com/ory/oathkeeper/issues/491

## Proposed changes

As described by RFC, the `aud` field from introspect could be a string or list of string. My propose is to create a custom type and handle the parse to avoid major changes on Oathkeeper.

## Checklist

- [X] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md).
- [X] I have read the [security policy](../security/policy).
- [X] I confirm that this pull request does not address a security
      vulnerability. If this pull request addresses a security. vulnerability, I
      confirm that I got green light (please contact
      [security@ory.sh](mailto:security@ory.sh)) from the maintainers to push
      the changes.
- [ ] I have added tests that prove my fix is effective or that my feature
      works.
- [ ] I have added or changed [the documentation](docs/docs).

## Further comments

Working on it. I will finish it tomorrow.

PR #822, opened 2021-09-16
chore: allow both string and []string in aud field

m1pl:
According to the [specification](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3), the `aud` field can either be an array of strings or a single string. Oathkeeper only supports arrays, which breaks the integration with IAM tools like Keycloak and forces users to use ugly hacks like setting an additional fake audience. This change adds support for strings to Oathkeeper.

## Related issue(s)

@aeneasr 
#491 
#601 
#792 
#810

## Checklist

- [X] I am following the
      the changes.

### DIFF
--- a/pipeline/authn/authenticator_oauth2_introspection.go
+++ b/pipeline/authn/authenticator_oauth2_introspection.go
@@ -82,18 +82,49 @@ func (a *AuthenticatorOAuth2Introspection) GetID() string {
 	return "oauth2_introspection"
 }
 
+type Audience []string
+
 type AuthenticatorOAuth2IntrospectionResult struct {
 	Active    bool                   `json:"active"`
 	Extra     map[string]interface{} `json:"ext"`
 	Subject   string                 `json:"sub,omitempty"`
 	Username  string                 `json:"username"`
-	Audience  []string               `json:"aud"`
+	Audience  Audience               `json:"aud,omitempty"`
 	TokenType string                 `json:"token_type"`
 	Issuer    string                 `json:"iss"`
 	ClientID  string                 `json:"client_id,omitempty"`
 	Scope     string                 `json:"scope,omitempty"`
 	Expires   int64                  `json:"exp"`
 	TokenUse  string                 `json:"token_use"`
+}
+
+func (a *Audience) UnmarshalJSON(b []byte) error {
+	var errUnsupportedType = errors.New("Unsupported aud type, only string or []string are allowed")
+
+	var jsonObject interface{}
+	err := json.Unmarshal(b, &jsonObject)
+	if err != nil {
+		return err
+	}
+
+	switch o := jsonObject.(type) {
+	case string:
+		*a = Audience{o}
+		return nil
+	case []interface{}:
+		s := make(Audience, 0, len(o))
+		for _, v := range o {
+			value, ok := v.(string)
+			if !ok {
+				return errUnsupportedType
+			}
+			s = append(s, value)
+		}
+		*a = s
+		return nil
+	}
+
+	return errUnsupportedType
 }
 
 func (a *AuthenticatorOAuth2Introspection) tokenFromCache(config *AuthenticatorOAuth2IntrospectionConfiguration, token string, ss fosite.ScopeStrategy) *AuthenticatorOAuth2IntrospectionResult {
--- a/pipeline/authn/authenticator_oauth2_introspection_test.go
+++ b/pipeline/authn/authenticator_oauth2_introspection_test.go
@@ -840,4 +840,22 @@ func TestAuthenticatorOAuth2Introspection(t *testing.T) {
 			require.NotEqual(t, noPreauthClient3, noPreauthClient)
 		})
 	})
+
+	t.Run("unmarshal-audience", func(t *testing.T) {
+		t.Run("Should pass because audience is a valid string", func(t *testing.T) {
+			var aud Audience
+			data := `"audience"`
+			json.Unmarshal([]byte(data), &aud)
+			require.NoError(t, err)
+			require.Equal(t, Audience{"audience"}, aud)
+		})
+
+		t.Run("Should pass because audience is a valid string array", func(t *testing.T) {
+			var aud Audience
+			data := `["audience1","audience2"]`
+			json.Unmarshal([]byte(data), &aud)
+			require.NoError(t, err)
+			require.Equal(t, Audience{"audience1", "audience2"}, aud)
+		})
+	})
 }


pipeline/authn/authenticator_oauth2_introspection.go
```diff
@@ -96,6 +98,35 @@ type AuthenticatorOAuth2IntrospectionResult struct {
 	TokenUse  string                 `json:"token_use"`
 }
 
+func (a *Audience) UnmarshalJSON(b []byte) error {
```
zepatrik, line 101: I mean the behavior is quite obvious, but please still add a test for this function to prevent regression issues when someone comes and changes that later on.

zepatrik (changes requested):
Thanks, the code itself looks good to me :+1:

pipeline/authn/authenticator_oauth2_introspection.go
m1pl: I'll add a test for both the string and the array case.

m1pl: @zepatrik I added some tests. What do you think?

pipeline/authn/authenticator_oauth2_introspection.go
```diff
@@ -96,6 +98,35 @@ type AuthenticatorOAuth2IntrospectionResult struct {
...
+	var errUnsupportedType = errors.New("Unsupported aud type, only string or []string are allowed")
+
+	var jsonObject interface{}
```
aeneasr, line 104: Since it could also be `[]string` I believe it might make sense to use gjson to assert the type and then decode into the right struct. WDYT?

aeneasr (changes requested):
Looking great :) As @zepatrik pointed out please add a test :)

pipeline/authn/authenticator_oauth2_introspection.go
m1pl: I don't think this is necessary, as I'm already checking for `[]string` below. No need to use gjson here.

m1pl:
@zepatrik @aeneasr I added tests.

pipeline/authn/authenticator_oauth2_introspection.go
zepatrik: :thinking: I only see cases for `[]interface{}` and `string`

aeneasr: As far as I see the code checks for `string` and `[]interface`. AFAIK it's possible that the JSON decoder also converts to `[]string`.

pipeline/authn/authenticator_oauth2_introspection_test.go
```diff
@@ -840,4 +840,22 @@ func TestAuthenticatorOAuth2Introspection(t *testing.T) {
...
 		})
 	})
+
+	t.Run("unmarshal-audience", func(t *testing.T) {
+		t.Run("Should pass because audience is a valid string", func(t *testing.T) {
+			var r AuthenticatorOAuth2IntrospectionResult
+			data := `{"aud":"audience"}`
+			json.Unmarshal([]byte(data), &r)
+			require.NoError(t, err)
+			require.Equal(t, Audience{"audience"}, r.Audience)
```
zepatrik, line 850: Wouldn't it make more sense to test the `Audience` type instead?
```suggestion
			var r Audience
			data := `"audience"`
			json.Unmarshal([]byte(data), &r)
			require.NoError(t, err)
			require.Equal(t, Audience{"audience"}, r)
```

pipeline/authn/authenticator_oauth2_introspection_test.go
```diff
@@ -840,4 +840,22 @@ func TestAuthenticatorOAuth2Introspection(t *testing.T) {
...
+		})
+
+		t.Run("Should pass because audience is a valid string array", func(t *testing.T) {
+			var r AuthenticatorOAuth2IntrospectionResult
+			data := `{"aud":["audience1","audience2"]}`
+			json.Unmarshal([]byte(data), &r)
+			require.NoError(t, err)
+			require.Equal(t, Audience{"audience1", "audience2"}, r.Audience)
```
zepatrik, line 858: same here

pipeline/authn/authenticator_oauth2_introspection_test.go
aeneasr: I think both approaches are fine. The current test ensures that the outer struct also correctly decodes, which is what we want here!

pipeline/authn/authenticator_oauth2_introspection_test.go
aeneasr: See https://github.com/ory/oathkeeper/pull/822#discussion_r712107876

pipeline/authn/authenticator_oauth2_introspection_test.go
m1pl: @zepatrik Would be easier, yes. I changed it. @aeneasr I don't think it makes any difference in this case. As the outer struct only containes the audience, it adds no value.

pipeline/authn/authenticator_oauth2_introspection.go
m1pl: @zepatrik @aeneasr Correct, the JSON decoder supports `[]string`. In this case we don't know whether it's `[]string` or `string` ahead of time, so we can't just decode into `[]string`. AFAIK this is the cleanest solution:
1. Unmarshal into generic `interface{}`
2. Check if type is `string`
3. Check if type is generic `[]interface{}` and build a `[]string`.

I tried to check for `[]string` instead of `[]interface{}`, it doesn't work. My Go skills are a little bit rusty as it has been a few years, although in this case I'm pretty confident it's a good solution. Do you have a better proposal? What don't you like about this solution?

aeneasr: I think the problem is that there are four possible cases:

1. `string`
1. `[]string`
1. `[]interface{}`

because `[]string` != `[]interface{}` - you need type conversion :)

So another solution might look like this:

```
var auds []string
if aud := gjson.Get(json, "aud"); aud.IsArray() {
  aud.ForEach(func(_, result) {
    auds = append(auds, result)
  })
} else if len(aud.Str) > 0 {
  auds = append(aud.Str)
} else {
  // error or something else
}
```

m1pl: Just for my understanding, can you provide an input where the current conversion fails to produce a valid output? Still don't see the problem.

m1pl: @aeneasr 
> because `[]string` != `[]interface{}` just catching `[]interface{}` is not enough - you need type conversion :)

https://github.com/ory/oathkeeper/pull/822/files#diff-5c8061290da17d352a4b888e5270cdd071ea584cd4751011132d8cccd69594f0R117

This looks like type conversion :) I'm just checking if it's a generic array, then cast all members to a string and fail if a member is not a string. The solution works without using external libraries like gjson. If you prefer to use gjson, feel free to edit the PR. If not, please consider merging it soon.

zepatrik: Go does not have subtyping rules, so `[]interface{}{} != []string{}`. What you liked is not a type conversion, but type assertion. Go will look whether the value is of the specified type, and `ok` is true if it is.
https://play.golang.org/p/DCbHNxY0JmJ gives you a type error although one is the sub type of the other.

But regardless, `json.Unmarshal` does not seem to give you a `[]string`: https://play.golang.org/p/MX9XbCzjCth so this code should be fine.

m1pl: @zepatrik @aeneasr Can you please merge this?

aeneasr: Oh, missed @zepatrik comment! Thank you :)

aeneasr (approved):

PR #835, opened 2021-09-24
docs: Warn that gzip is unsupported

danrods:
Note to users that gzip responses are as of now unsupported for Cookie and Bearer authenticators. 
The result is that the `subject` and `extra` will not be filled in, and will fail silently.

## Checklist

- [x] I am following the
      the changes.
      works.

### DIFF
--- a/docs/docs/pipeline/authn.md
+++ b/docs/docs/pipeline/authn.md
@@ -234,7 +234,8 @@ authenticator is enabled for this URL.
 The `cookie_session` authenticator will forward the request method, path and
 headers to a session store. If the session store returns `200 OK` and body
 `{ "subject": "...", "extra": {} }` then the authenticator will set the subject
-appropriately.
+appropriately. Please note that Gzipped responses from `check_session_url` are 
+not currently supported, and will fail silently.
 
 ### Configuration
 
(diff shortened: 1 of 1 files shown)


aeneasr (changes requested):
Golang handles gziped HTTP response bodies automatically so this is supported. Maybe your response is missing the appropriate content encoding headers (and maybe also not respecting the request headers although that could be a bug in our forward logic).

Can you share your HTTP Response Headers?

danrods:
@aeneasr Thank you for the review

I would have assumed so as well. This stemmed from a discussion we had in the oathkeeper slack channel
https://ory-community.slack.com/archives/C01340V8KSM/p1632166348046700

In this thread you can see the response headers I was sending
https://ory-community.slack.com/archives/C01340V8KSM/p1632237741052700?thread_ts=1632213654.049300&cid=C01340V8KSM

It contains the proper `Content-Encoding` header for `gzip`. We've been using this gzip compression filter in production for a long time and have never had issues. The only thing I know is that when the response is compressed the `AuthenticationSession` is not filled in with the body of the response, but after I removed the compression it is. 

I wanted to update the docs because theres no indication that there was a failure, I just couldn't print out the subject

aeneasr:
Hm strange, thank you in that case! I have opened an issue to track this :)

danrods:
Sounds good! Should I close this PR then, or should we add to the docs temporarily until it can be addressed? @aeneasr

aeneasr:
I'll merge this one so people can at least troubleshoot :) Thank you! :)

PR #832, opened 2021-09-23
feat: Integrate Zipkin tracer

tobbbles:
Updates the `ory/x/tracing` package to the latest version, configures the provider to parse config file and envvar configuration for Zipkin, integrates the tracer

## Checklist

- [X] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md).
- [X] I am following the
      [contributing code guidelines](../blob/master/CONTRIBUTING.md#contributing-code).
- [X] I have read the [security policy](../security/policy).
- [X] I confirm that this pull request does not address a security
      vulnerability. If this pull request addresses a security. vulnerability, I
      confirm that I got green light (please contact
      [security@ory.sh](mailto:security@ory.sh)) from the maintainers to push
      the changes.
- [ ] I have added tests that prove my fix is effective or that my feature
      works.
- [X] I have added or changed [the documentation](docs/docs).

## Further Comments

The `go.mod` `replace` directive for updating just the tracing package is horrible, but there is no straight upgrade path for the `ory/x` package, since it still relies on the viperx package. I'd be happy to discuss the semantics around this and oathkeeper

PR #838, opened 2021-09-29
fix: Making docs config module aware for github.com/ory/x

tobbbles:
Currently the master branch is failing due to #832 being merged, and the ory/x tag being upgraded. This docs gen failure was previously attempted to be fixed in https://github.com/ory/oathkeeper/commit/71b49f846bc4cfff8551b9bec832255adbc78afd by @zepatrik but including a Go package at the root of `github.com/ory/x`, however this was only for the one tag and isn't in the mainline.

This fix addresses the root cause (I hope) by making the `go list` list the modules of the repo rather than the global package

## Related issue(s)
#803

## Checklist

- [X] I am following the
      the changes.
      works.

### DIFF
--- a/docs/scripts/config.js
+++ b/docs/scripts/config.js
@@ -11,7 +11,7 @@ const { execSync } = require('child_process')
 const fetch = require('node-fetch')
 
 const oryXVersion = execSync(
-  "cd ..; go list -f '{{.Module.Version}}' -find github.com/ory/x"
+  "cd ..; go list -f '{{.Version}}' -m github.com/ory/x"
 )
   .toString('utf-8')
   .trim()


docs/scripts/config.js
```diff
@@ -11,7 +11,7 @@ const { execSync } = require('child_process')
 const fetch = require('node-fetch')
 
 const oryXVersion = execSync(
-  "cd ..; go list -f '{{.Module.Version}}' -find github.com/ory/x"
```
aeneasr, line 14: Unfortunately this will be overwritten if we change the docs template. What is the underlying issue here? Is it the go version?

tobbbles: Can we also update the template?

The issue is that the existing `go list` without a `-m` flag isn't module aware, when it tries to list the packages this way the following happens:
```sh
$ go list -f '{{.Module.Version}}' github.com/ory/x
package github.com/ory/x: build constraints exclude all Go files in /Users/tobbbles/code/go/pkg/mod/github.com/ory/x@v0.0.165
```

This appears to be due to github.com/ory/x not containing any Go files in the root, so `go list` doesn't deem it a valid package.

This was previously solved by just touching a go package in the ory/x root on a 'patch' branch; but IMO the issue is this incorrect listing from `go list` not resolving the _module_ version, but trying to resolve the used package version

aeneasr: @zepatrik wrote this - can you chime in?

tobbbles: I see https://github.com/ory/x/blob/master/package.go was added 8 months ago and that's why projects on newer ory/x packages don't receive this error, however we can't upgrade oathkeeper yet due to the viperx->koanf requirement.

fwiw, the module aware go list will also works in kratos on the newer version of ory/x

```sh
$ go list -f '{{.Version}}' -m github.com/ory/x
v0.0.272

$ go list -f '{{.Module.Version}}' github.com/ory/x
v0.0.272
```

zepatrik: Ah ok nice, then we can use it with `-m`. :+1:
The right place to update this script is here: https://github.com/ory/docusaurus-template/blob/8ee6f932f6ef3edf5c471bff4eb64dcbb9d22a0c/scripts/config.js#L14

aeneasr:
Merged https://github.com/ory/docusaurus-template/pull/79

PR #882, opened 2021-12-13
feat: Enable custom cache key support for Hydrator

anderslauri:
The cache key for `hydrator` is composed of the whole `AuthenticationSession` - this is not optimal. This PR provides the ability to opt-in to a special cache key as (as part of) `AuthenticationSession` for more fine granular cache control

## Related issue(s)
- None what I found.
## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/.schema/config.schema.json
+++ b/.schema/config.schema.json
@@ -1127,6 +1127,11 @@
             "enabled": {
               "$ref": "#/definitions/handlerSwitch"
             },
+            "key": {
+              "type": "string",
+              "title": "Cache key",
+              "description": "Define custom cache key, i.e '{{ print .Subject }}'. All properties must be part of AuthenticationSession."
+            },
             "ttl": {
               "type": "string",
               "pattern": "^[0-9]+(ns|us|ms|s|m|h)$",
--- a/go.mod
+++ b/go.mod
@@ -25,7 +25,7 @@ require (
 	github.com/go-openapi/strfmt v0.19.5
 	github.com/go-openapi/swag v0.19.9
 	github.com/go-sql-driver/mysql v1.5.0
-	github.com/go-swagger/go-swagger v0.25.0
+	github.com/go-swagger/go-swagger v0.28.0
 	github.com/gobuffalo/httptest v1.0.2
 	github.com/gobuffalo/packr/v2 v2.8.0
 	github.com/gobwas/glob v0.2.3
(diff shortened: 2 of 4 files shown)


anderslauri:
Not sure why tests are failling, looks like a `swagger` issue.

anderslauri:
OK, now some bash-script for E2E-tests are failing, `killall` is not found also some file generation or a path is incorect.

anderslauri:
🤔 

Looks like there is no actual coverage for the cache by the existing tests.

aeneasr (changes requested):
Not sure if this is intentional but this would use exactly one cache key per route for all requests. Using custom cache keys is extremely dangerous because it can lead to results which say someone is allowed to do something while in fact they are not.

We had this problem already in Oathkeeper which is why i am very reluctant to merge changes to the cache behavior. I hope you understand!

If I misunderstood the use case here please LMK :)

anderslauri:
> We had this problem

Maybe I missunderstood the flexible configuration. Let us be transparent - if one chooses to have a static cache key, naturally as you describe the situation would be correct. However I was under the assumption one can do, i.e. `{{ print .Subject }}` and thus that would be the individual subject on the `AuthenticationSession` - if this is shared or not - in the end it is in the hand of the consumer. However, given this flexiblity provides a more custom approach to set a cache key beyond the whole `AuthenticationSession`.

aeneasr:
The value `cfg.Cache.Key` is static globally, and can be overwritten per route. Payloads or sessions have no effect on it. You would thus always get the same result for any request to the specified route(s), regardless of your payloads.

anderslauri:
Thank you,

Then my fundamental assumption was incorrect! PR as-is now have little
value, let me give this some thought as we would require a bit of
flexibility here.

aeneasr:
Thank you for contributing :) In that case, I'll close this PR and am looking forward to your next contribution :)

anderslauri:
> The value `cfg.Cache.Key` is static globally, and can be overwritten per route. Payloads or sessions have no effect on it. You would thus always get the same result for any request to the specified route(s), regardless of your payloads.

Following up here. OK, I understand the issue you are finally describing. I will try to rephrase this for myself.

1. `Subject A` is using `Route A` proxied through oathkeeper where one part of the pipeline is `hydrator`, this response is cached with the key `Subject A`. The entire `AuthenticationSession` is cached.
2. `Subject A` is using a another system and is routed using `Route B`, as the cache key is `print .subject` the `AuthenticationSession` from `Route A` would be the response and thus could include bad data.

Assuming the above is correct. A cache key can still be composed to be uniquely enough to work, however, yes it can be fragile. The follow up question would be, as the layout is today within the code, the pipeline (a defined set of routes) - does this have an unique identifier which can be accessed? If so the one can form a composite key (cache key + pipeline|route id) - assuming such exists?

Issue #515, opened 2020-09-14
remote_json: Enable timeout configuration for calls to authorization endpoint


Sbou:
**Is your feature request related to a problem? Please describe.**

When running on Kubernetes with remote_json we find that the call to the authorization endpoint often times out. The default timeout is 500 milliseconds, which is a long time and should be enough but we find that for our setup, it generally isn't.

**Describe the solution you'd like**
I'd like to be able to specify a timeout for the request to the authorization endpoint so that we are not hindered by things that really should be circumventable. Something akin to the retry config that you can set for authorization should do the trick.

**Describe alternatives you've considered**

**Additional context**

aeneasr:
Sounds good, contributions appreciated!

Kortivex:
Can be this applies to the 'remote' authorizer too?

aeneasr:
Yes!

PR #883, opened 2021-12-16
feat: Add retry and timeout support in authorizers

Sbou:
## Related issue(s)

As for authenticators, it is possible to define a retry and timeout policy on remote authorizers. 
This need was expressed in the issue #515.

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/pipeline/authz/remote.go
+++ b/pipeline/authz/remote.go
@@ -44,8 +44,9 @@ type AuthorizerRemote struct {
 // NewAuthorizerRemote creates a new AuthorizerRemote.
 func NewAuthorizerRemote(c configuration.Provider) *AuthorizerRemote {
 	return &AuthorizerRemote{
-		c: c,
-		t: x.NewTemplate("remote"),
+		c:      c,
+		client: httpx.NewResilientClientLatencyToleranceSmall(nil),
+		t:      x.NewTemplate("remote"),
 	}
 }
 
--- a/pipeline/authz/remote.go
+++ b/pipeline/authz/remote.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"net/http"
 	"text/template"
+	"time"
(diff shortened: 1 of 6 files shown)


pipeline/authz/remote_json.go
```diff
@@ -42,9 +49,8 @@ type AuthorizerRemoteJSON struct {
 // NewAuthorizerRemoteJSON creates a new AuthorizerRemoteJSON.
 func NewAuthorizerRemoteJSON(c configuration.Provider) *AuthorizerRemoteJSON {
 	return &AuthorizerRemoteJSON{
-		c:      c,
-		client: httpx.NewResilientClientLatencyToleranceSmall(nil),
```
aeneasr, line 46: Is it possible that this could cause a panic? I think we should leave the value initialized.

pipeline/authz/remote.go
```diff
@@ -37,9 +44,8 @@ type AuthorizerRemote struct {
 // NewAuthorizerRemote creates a new AuthorizerRemote.
 func NewAuthorizerRemote(c configuration.Provider) *AuthorizerRemote {
 	return &AuthorizerRemote{
-		c:      c,
-		client: httpx.NewResilientClientLatencyToleranceSmall(nil),
```
aeneasr, line 41: Is it possible that this could cause a panic? I think we should leave the value initialized.

aeneasr (changes requested):

pipeline/authz/remote_json.go
Sbou: Ok, i'm going to put back this line.
I had copied the behavior implemented in the Oauth2 authenticator.

aeneasr:
Thank you very much!

PR #885, opened 2021-12-17
feat: Composite configurable mutator cache key per rule.

anderslauri:
Enable support for configurable mutator key for unique rule.

## Related issue(s)
None known.

## Checklist

- [x ] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md).
      introduces a new feature.
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/.schema/config.schema.json
+++ b/.schema/config.schema.json
@@ -1127,6 +1127,11 @@
             "enabled": {
               "$ref": "#/definitions/handlerSwitch"
             },
+            "key": {
+              "type": "string",
+              "title": "Cache key",
+              "description": "Define custom cache key, i.e '{{ print .Subject }}'. All properties must be part of AuthenticationSession."
+            },
             "ttl": {
               "type": "string",
               "pattern": "^[0-9]+(ns|us|ms|s|m|h)$",
--- a/pipeline/mutate/mutator_hydrator.go
+++ b/pipeline/mutate/mutator_hydrator.go
@@ -241,8 +241,11 @@ func (a *MutatorHydrator) Mutate(r *http.Request, session *authn.AuthenticationS
 	}
 	*session = sessionFromUpstream
 
-	a.hydrateToCache(cfg, encodedSession, session)
-
+	if len(cfg.Cache.Key) > 0 {
+		a.hydrateToCache(cfg, cfg.Cache.Key, session)
+	} else {
+		a.hydrateToCache(cfg, encodedSession, session)
+	}
 	return nil
 }
 
--- a/pipeline/mutate/mutator_hydrator.go
+++ b/pipeline/mutate/mutator_hydrator.go
@@ -141,18 +141,19 @@ func (a *MutatorHydrator) hydrateToCache(config *MutatorHydratorConfig, key stri
 	}
 }
 
-func (a *MutatorHydrator) Mutate(r *http.Request, session *authn.AuthenticationSession, config json.RawMessage, _ pipeline.Rule) error {
+func (a *MutatorHydrator) Mutate(r *http.Request, session *authn.AuthenticationSession, config json.RawMessage, p pipeline.Rule) error {
 	cfg, err := a.Config(config)
 	if err != nil {
 		return err
 	}
 
 	if len(cfg.Cache.Key) > 0 {
-		if cacheSession, ok := a.hydrateFromCache(cfg, cfg.Cache.Key); ok {
+		if cacheSession, ok := a.hydrateFromCache(cfg, cfg.Cache.Key + p.GetID()); ok {
 			*session = *cacheSession
 			return nil
 		}
-		a.d.Logger().Debugf("Cache key %s was not found. Falling back on default.", cfg.Cache.Key)
+		a.d.Logger().Debugf("Cache key %s for rule %s id was not found. Falling back on default.",
+			cfg.Cache.Key, p.GetID())
 	}
 	var b bytes.Buffer
 	if err := json.NewEncoder(&b).Encode(session); err != nil {
@@ -242,7 +243,7 @@ func (a *MutatorHydrator) Mutate(r *http.Request, session *authn.AuthenticationS
 	*session = sessionFromUpstream
 
 	if len(cfg.Cache.Key) > 0 {
-		a.hydrateToCache(cfg, cfg.Cache.Key, session)
+		a.hydrateToCache(cfg, cfg.Cache.Key + p.GetID(), session)
 	} else {
 		a.hydrateToCache(cfg, encodedSession, session)
 	}
--- a/pipeline/mutate/mutator_hydrator.go
+++ b/pipeline/mutate/mutator_hydrator.go
@@ -85,7 +85,7 @@ type externalAPIConfig struct {
 type cacheConfig struct {
 	Enabled bool   `json:"enabled"`
 	TTL     string `json:"ttl"`
-
+	Key     string `json:"key"`
 	ttl time.Duration
 }
 
@@ -147,6 +147,13 @@ func (a *MutatorHydrator) Mutate(r *http.Request, session *authn.AuthenticationS
 		return err
 	}
 
+	if len(cfg.Cache.Key) > 0 {
+		if cacheSession, ok := a.hydrateFromCache(cfg, cfg.Cache.Key); ok {
+			*session = *cacheSession
+			return nil
+		}
+		a.d.Logger().Debugf("Cache key %s was not found. Falling back on default.", cfg.Cache.Key)
+	}
 	var b bytes.Buffer
 	if err := json.NewEncoder(&b).Encode(session); err != nil {
 		return errors.WithStack(err)
--- a/pipeline/mutate/mutator_hydrator.go
+++ b/pipeline/mutate/mutator_hydrator.go
@@ -86,7 +86,7 @@ type cacheConfig struct {
 	Enabled bool   `json:"enabled"`
 	TTL     string `json:"ttl"`
 	Key     string `json:"key"`
-	ttl time.Duration
+	ttl     time.Duration
 }
 
 type MutatorHydratorConfig struct {
--- a/pipeline/mutate/mutator_hydrator_test.go
+++ b/pipeline/mutate/mutator_hydrator_test.go
@@ -127,6 +127,14 @@ func configWithRetriesForMutator(giveUpAfter, retryDelay string) func(*httptest.
 	}
 }
 
+func configWithSpecialCacheKey(key string) func(*httptest.Server) json.RawMessage {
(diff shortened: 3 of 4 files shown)


pipeline/mutate/mutator_hydrator.go
```diff
@@ -114,16 +114,16 @@ func (a *MutatorHydrator) GetID() string {
 	return "hydrator"
 }
 
-func (a *MutatorHydrator) cacheKey(config *MutatorHydratorConfig, session string) string {
-	return fmt.Sprintf("%s|%x", config.Api.URL, md5.Sum([]byte(session)))
+func (a *MutatorHydrator) cacheKey(config *MutatorHydratorConfig, key string) string {
+	return fmt.Sprintf("%s|%x", config.Api.URL, md5.Sum([]byte(key)))
```
aeneasr, line 118: The key is taken as a literal and not being executed. So `{{ print .Subject }}` would just be interpreted as a literal string, not the actual subject ID :)

If you want to execute it, you need to render the template and give it the data.

pipeline/mutate/mutator_hydrator_test.go
```diff
@@ -346,6 +353,15 @@ func TestMutatorHydrator(t *testing.T) {
 				Match:   newAuthenticationSession(),
 				Err:     nil,
 			},
+			"Custom Cache Key": {
+				Setup:   defaultRouterSetup(),
+				Session: newAuthenticationSession(setSubject(sampleSubject)),
+				Rule:    &rule.Rule{ID: "test-rule"},
+				Config:  configWithSpecialCacheKey(sampleSubject),
```
aeneasr, line 360: According to the documentation, this should work with substitution / templating. It would be good if you could add some tests that cover this, and also ensure that no cache key is the same when templating is executed appropriately.

.schema/config.schema.json
```diff
@@ -1127,6 +1127,11 @@
             "enabled": {
               "$ref": "#/definitions/handlerSwitch"
             },
+            "key": {
+              "type": "string",
+              "title": "Cache key",
+              "description": "Define custom cache key, i.e '{{ print .Subject }}'. All properties must be part of AuthenticationSession."
```
aeneasr, line 1133: One difficulty with this is that a singular typo (e.g. `{{ print .Subject}` instead of `{{ print .Subject }}`) can lead to catastrophic cache confusion. So it's very easy to get this wrong because we don't know if this is a literal string or a template string.

aeneasr (changes requested):

.schema/config.schema.json
anderslauri: Let me give this some further thought what can be done here while retaining flexibility. As of now the following assertions are in place to reduce likelyhood of cache confusion.

1. `subject` must match cached `subject` of `AuthenticationSession`.
2. Configured URL of `hydrator` is part of the composite cache key, i.e. no cache key of another configured `hydrator` URL will overlap.
3. Rule ID will be part of the of the composite cache key (as per this PR), ensuring cache key scope is limited to rule.

pipeline/mutate/mutator_hydrator.go
```diff
@@ -107,68 +110,83 @@ func NewMutatorHydrator(c configuration.Provider, d mutatorHydratorDependencies)
 		// This is a best-practice value.
 		BufferItems: 64,
 	})
-	return &MutatorHydrator{c: c, d: d, client: httpx.NewResilientClientLatencyToleranceSmall(nil), hydrateCache: cache}
+	return &MutatorHydrator{c: c, d: d,
+		p: sync.Pool{
```
anderslauri, line 114: Introduction of `sync.Pool` make sense here to re-use `bytes.Buffer` when possible.

pipeline/mutate/mutator_hydrator.go
```diff
@@ -107,68 +110,83 @@ func NewMutatorHydrator(c configuration.Provider, d mutatorHydratorDependencies)
...
-		return nil
+	err = json.NewEncoder(sessionCacheKey).Encode(session)
+	switch {
+	case err != nil:
+		return errors.WithStack(err)
+	case !cfg.Cache.Enabled:
+		break
+	case len(cfg.Cache.Key) > 0:
+		// Build a composite cache key with property from configuration.
+		if cacheSession, ok := a.hydrateFromCache(a.cacheKey(
```
anderslauri, line 167: Key is a composite key based on configuration property + three distinct repeatable properties.

pipeline/mutate/mutator_hydrator.go
anderslauri: Done.

.schema/config.schema.json
anderslauri: I believe the existing method is as accurate as it can be.

pipeline/mutate/mutator_hydrator_test.go
anderslauri: Done.

anderslauri:
> Awesome, thank you for your contribution! This looks pretty good and I have some ideas how to improve it further :)

Please have a new review. Should be fine.

> Awesome, thank you for your contribution! This looks pretty good and I have some ideas how to improve it further :)

Thank you, please have another review. This includes some refactoring, introduction of `sync.Pool` and increased test-coverage of cache, both for custom key and also default behavior.

pipeline/mutate/mutator_hydrator.go
aeneasr: I am not sure if that is enough! What about extra properties of the session, such as e.g. "scope" or "permissions"? These could all influence the hydrator response and could lead to eventual security vulnerabilities down the road. I think we need to take the full session in a serialized form if we want to be sure that the cache can actually be reused!

pipeline/mutate/mutator_hydrator.go
aeneasr: I would propose to try and not reuse buffers. It can lead to multiple requests re-using buffers and breaking boundaries between requests. While this might not be easily achievable, it can be an attack vector (think side channel attacks).

pipeline/mutate/mutator_hydrator.go
```diff
@@ -102,73 +104,86 @@ func NewMutatorHydrator(c configuration.Provider, d mutatorHydratorDependencies)
...
 
-	var b bytes.Buffer
-	if err := json.NewEncoder(&b).Encode(session); err != nil {
-		return errors.WithStack(err)
-	}
+	s := a.p.Get().(*bytes.Buffer)
+	defer func() {
+		s.Reset()
+		a.p.Put(s)
+	}()
```
aeneasr, line 156: I would propose to keep this PR focused on making caching possible for mutators. Any other optimizations should be done in a separate PR :)

pipeline/mutate/mutator_hydrator_test.go
```diff
@@ -346,27 +406,85 @@ func TestMutatorHydrator(t *testing.T) {
...
+					// a good strategy, however, the alternative would be to replace the cache.
+					// See https://github.com/dgraph-io/ristretto/blob/9d4946d9b973c8e860ae42944e07f5bbe28a506b/cache_test.go#L17
+					time.Sleep(cacheTestWriteSleep)
+				case testName == cacheTestAuthSessionName:
+					// Ensure session is persisted within cache.
+					require.NoError(t, a.Mutate(specs.Request, specs.Session, specs.Config(ts), specs.Rule))
+					time.Sleep(cacheTestWriteSleep)
+				}
+
+				if err := a.Mutate(specs.Request, specs.Session, specs.Config(ts), specs.Rule); specs.Err == nil {
```
aeneasr, line 470: These tests look much better! Another option you have would be to add setup and assert functions to the `testMap` and then run them here if they are not nil. That makes the test setup easier to read and follow. An alternative is also to write a dedicated test for this, you don't necessarily have to reuse the code that exists. Sometimes explicit code is better, even if there is a bit of duplication involved :)

aeneasr (changes requested):

pipeline/mutate/mutator_hydrator.go
anderslauri: OK, sure, I will remove this. Even though we already know the variable will be allocated on the heap due to passing of a pointer to json decode.

pipeline/mutate/mutator_hydrator.go
anderslauri: If we take the full `AuthenticationSession` we're not really benefiting anything given the volatiltity of this structure - using `subject`, `rule id`, `hydrator URL` as part of the composition is about enabling flexibility while reducing chance of cache leakage. We also need to remember this is opt-in and not opt-out as part of configuration.

pipeline/mutate/mutator_hydrator_test.go
anderslauri: Done.

pipeline/mutate/mutator_hydrator.go
aeneasr: Thank you. I am unsure though wether I can agree here. I think it points to a deeper issue though which is manifesting in this pr: It is not possible to define the cache key reliably. Actually, one would probably like to define the cache key themselves (e.g. subject + scope) to make the system more efficient. For example, the AuthSession might contain vital info such as a "permissions" array. Yet, it may also contain a counter or timestamp which is changing for most of the requests - invalidating the cache if included.

We are still not there yet with a new concept for Ory Oathkeeper, but I think this is a very interesting problem that could very well warrant a realignment on Ory Oathkeeper which would be to make access control at the reverse proxy as efficient and flexible as possible.

I think this too could benefit from JsonNet, as JsonNet is typable, lintable, and can produce errors (go templating fulfills none of these properties).

Unfortunately, for this PR in particular, I don't think the current implementation can be accepted because it still bears too many risks and this particular type of issue has already caused a CVE in Ory Oathkeeper which is why I am so hyper-sensible about this topic: https://github.com/ory/oathkeeper/security/advisories/GHSA-qvp4-rpmr-xwrr

aeneasr (changes requested):

pipeline/mutate/mutator_hydrator.go
anderslauri: Thank you. Let us inverse the solution. If you feel a user can't define a cache key freely (given cause of concerns as mentioned) - what about allowing the user to define exclusion of request/response `headers`? If the whole `AuthenticationSession` is used as a cache key `headers` represents a volatile part. `extra` is defined as part of previous steps in the pipeline and can be more controlled. What is your thoughts?

aeneasr: That's a brilliant idea! Exclusion is much safer (because explicit) :)

Issue #521, opened 2020-09-24
Decision API support for the nginx-ingress annotation nginx.ingress.kubernetes.io/auth-url 


AlbinoDrought:
**Is your feature request related to a problem? Please describe.**

We are using `nginx-ingress` with the `nginx.ingress.kubernetes.io/auth-url` annotation.

We want to apply an access rule to this URL: https://admin.site.example/dashboard

We have an ingress config that looks like this:

```yaml
ingress:
  enabled: true
  annotations:
    kubernetes.io/ingress.class: nginx
    nginx.ingress.kubernetes.io/auth-url: "http://oathkeeper.site.example/decisions$request_uri"
    nginx.ingress.kubernetes.io/auth-response-headers: Authorization
```

Oathkeeper expects (correct me if I'm wrong) a request that looks like this:

```
GET /decisions/dashboard HTTP/2
Host: admin.site.example
```

Oathkeeper will actually receive:

```
GET /decisions/dashboard HTTP/2
Host: oathkeeper.site.example
```

[The default nginx-ingress template](https://github.com/kubernetes/ingress-nginx/blob/3c8817f700a4ab1713e3369fc6e5f500b008d989/rootfs/etc/nginx/template/nginx.tmpl#L977) generates an NGINX config where the `Host` header sent to Oathkeeper is the auth_request host `oathkeeper.site.example` instead of our target host `admin.site.example`:

<details>
<summary>Generated nginx-ingress config</summary>

```
	## start server admin.site.example
	server {
		server_name admin.site.example ;		
		listen 80 proxy_protocol ;
		listen 443 proxy_protocol proxy_protocol ssl http2 ;		
		set $proxy_upstream_name "-";		

		ssl_certificate_by_lua_block {
			certificate.call()
		}		

		location = /_external-auth-Lw {
			internal;			# ngx_auth_request module overrides variables in the parent request,
			# therefore we have to explicitly set this variable again so that when the parent request
			# resumes it has the correct value set for this variable so that Lua can pick backend correctly
			set $proxy_upstream_name "example-site-admin-web-80";			
			proxy_pass_request_body     off;
			proxy_set_header            Content-Length          "";
			proxy_set_header            X-Forwarded-Proto       "";
			proxy_set_header            X-Request-ID            $req_id;			
			proxy_set_header            Host                    oathkeeper.site.example;
			proxy_set_header            X-Original-URL          $scheme://$http_host$request_uri;
[...]

PR #904, opened 2022-02-01
feat: Traefik decision api support

dadrus:
@aeneasr: This is the PR I was talking to you about yesterday

## Related issue(s)
This PR focuses on the decision API only. So I don't know, whether other traefik specific use cases are supported.

There are a couple of discussions, issues and stale PRs related to this PR:
* https://github.com/ory/oathkeeper/issues/521
* https://github.com/ory/oathkeeper/issues/441
* https://github.com/ory/oathkeeper/pull/487/
* https://github.com/ory/oathkeeper/issues/263
* https://github.com/ory/oathkeeper/discussions/899

## Checklist
- [x] I have referenced an issue containing the design document if my change introduces a new feature.
- [x] I am following the [contributing code guidelines](../blob/master/CONTRIBUTING.md#contributing-code).
- [x] I confirm that this pull request does not address a security vulnerability. If this pull request addresses a security. vulnerability, I confirm that I got green light (please contact [security@ory.sh](mailto:security@ory.sh)) from the maintainers to push the changes.
- [x] I have added tests that prove my fix is effective or that my feature works.

## Further Comments

Even there are a couple of ideas outlined in the referenced issues/PRs on how oathkeeper could be enhanced to have a more generic implementation to support different use cases (like other API gateways or proxies), this PR does not take them into account. Rather is focuses on Traefik support only. I'm convinced, it is however a good starting point to move forward and to resolve at least a couple of requests here at github, as well as at Slack. Further enhancements can then be done in further iterations. I also didn't want to touch any parts of the configuration, as we should first make use of the updated ory-x configuration capabilities.

Tests and documentation will be added as soon as we have a common understanding on how to proceed.

Beyond that, the given PR enables full support for the Traefik's [ForwardAuth](https://doc.traefik.io/traefik/middlewares/http/forwardauth/) middleware.

Please note that this implementation does not differentiate between different strategies (where to get the information about the upstream service uri, host, scheme and method). If no `X-Forwarded-*` header is set for a particular attribute, it is taken from the Request URL, otherwise the corresponding value from the header is taken. This way one can overwrite any, or none of the aforesaid attributes which is a non-breaking change to the current behavior.

I have not implemented a separate decision end-point by intention, as it would uselessly blow the code base by mainly copying&pasting code (implementation & tests) and make the project less maintainable. I'm also not really sure, whether an introduction of different strategies makes sense. But if it is according to the community, I personally favor a strategy based approach (via configuration) much more (which depends on the new ory-x version), compared to specific endpoints.


dduzgun-security:
This PR would be a solution for the discussion https://github.com/ory/oathkeeper/discussions/899 for the integration of Traefik and Oathkeeper

aeneasr (changes requested):
Looking solid, just needs some tests :)

dadrus:
@aeneasr: In principle, this PR is complete. I also added traefik to the list of supported integrations. The documentation is however kind of thin. If there would be a section for integration examples, I would add some. I'm also not sure, where to put the information about how the url to be tested is created by oathkeeper (request url  + header combinations)

dadrus:
@aeneasr: I'm not sure, whether the new guide about integrating oathkeeper  with traefik is properly referenced in the `sidebar.json`. Please take a look.

credentials/verifier_default.go
```diff
@@ -42,7 +43,7 @@ func (v *VerifierDefault) Verify(
 
 		kid, ok := token.Header["kid"].(string)
 		if !ok || kid == "" {
-			return nil, errors.WithStack(herodot.ErrInternalServerError.WithReason("The JSON Web Token must contain a kid header value but did not."))
+			return nil, errors.WithStack(herodot.ErrBadRequest.WithReason("The JSON Web Token must contain a kid header value but did not."))
```
aeneasr, line 46: Changing all of these is kind of a breaking change (although not an terrible one). Given that our service configuration returns something incorrect here, why would we report to the client calling this endpoint that it's a bad request?

aeneasr (changes requested):
Thank you! Great stuff :)

Regarding the docs, we are currently in the process of moving all the docs to github.com/ory/docs to resolve several major problems in the CI infrastructure. You can either keep them in this branch and I merge them into ory/docs once this is merged, or we make a PR against this PR ( https://github.com/ory/docs/pull/557 ) in this directory: https://github.com/ory/docs/tree/next-generation/docs/oathkeeper

dadrus:
Let's do a PR against the ory/docs PR.

credentials/verifier_default.go
dadrus: According to my understanding of the code, this is actually not about our service configuration, it is about the data sent by the client. If the client sent a malformed request (here the an access token not containing the expected data), it should imho be answered accordingly.

Another reason was logging. Internal server error is about unrecoverable errors, which imply some kind of a bug, which should be addressed. Here, we don't have things like this. So I wanted avoiding the be filled with error log statements, which are no errors.

Do you agree, or do you still want to have these changes reverted?

aeneasr: I see, in case that we verify credentials, `Unauthorized` error would also be acceptable. What do you think?

dadrus:
The last commit (see above) includes the merge with current master and removes the documentation. Latter is moved to the other PR (also referenced above)

credentials/verifier_default.go
dadrus: Please let me know if this is what you would like to have. I'll then update the PR accordingly.

aeneasr (approved):

designermonkey:
I just spotted the `v0.38.20-beta.1` tag.

PR #916, opened 2022-02-09
debug build

till:
- Fix: response lost
- Chore: include tag/commit (instead of master)
- Fix: gzip handling

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,7 +79,8 @@ install: .bin/packr2
 .PHONY: docker
 docker: .bin/packr2
 		packr2 || (GO111MODULE=on go install github.com/gobuffalo/packr/v2/packr2 && packr2)
-		CGO_ENABLED=0 GO111MODULE=on GOOS=linux GOARCH=amd64 go build
+		CGO_ENABLED=0 GO111MODULE=on GOOS=linux GOARCH=amd64 go build \
+			-ldflags "-X github.com/ory/oathkeeper/x.Version=$$(git describe)"
 		packr2 clean
 		docker build -t oryd/oathkeeper:dev .
 		docker build -t oryd/oathkeeper:dev-alpine -f Dockerfile-alpine .
--- a/pipeline/authn/authenticator_bearer_token.go
+++ b/pipeline/authn/authenticator_bearer_token.go
@@ -107,6 +107,10 @@ func (a *AuthenticatorBearerToken) Authenticate(r *http.Request, session *Authen
 		return helper.ErrForbidden.WithReasonf("The configured extra_from GJSON path returned an error on JSON output: %s", err.Error()).WithDebugf("GJSON path: %s\nBody: %s\nResult: %s", cf.ExtraFrom, body, extraRaw).WithTrace(err)
 	}
 
+	if len(subject) == 0 {
+		return helper.ErrForbidden.WithReasonf("Could not decode subject from response.")
+	}
+
 	session.Subject = subject
 	session.Extra = extra
 	return nil
--- a/pipeline/authn/authenticator_cookie_session.go
+++ b/pipeline/authn/authenticator_cookie_session.go
@@ -155,6 +155,10 @@ func forwardRequestToSessionStore(r *http.Request, checkSessionURL string, prese
 
 	// We need to make a COPY of the header, not modify r.Header!
 	for k, v := range r.Header {
+		// remove Accept-Encoding to let the transport handle gzip
+		if k == "Accept-Encoding" {
+			continue
+		}
 		req.Header[k] = v
 	}
 
--- a/pipeline/authn/authenticator_cookie_session.go
+++ b/pipeline/authn/authenticator_cookie_session.go
@@ -171,13 +171,14 @@ func forwardRequestToSessionStore(r *http.Request, checkSessionURL string, prese
 		return nil, helper.ErrForbidden.WithReason(err.Error()).WithTrace(err)
 	}
 
-	if res.StatusCode == 200 {
-		body, err := ioutil.ReadAll(res.Body)
-		if err != nil {
-			return json.RawMessage{}, errors.WithStack(herodot.ErrInternalServerError.WithReasonf("Unable to fetch cookie session context from remote: %+v", err))
-		}
-		return body, nil
-	} else {
-		return json.RawMessage{}, errors.WithStack(helper.ErrUnauthorized)
+	body, err := ioutil.ReadAll(res.Body)
+	if err != nil {
+		return json.RawMessage{}, errors.WithStack(herodot.ErrInternalServerError.WithReasonf("Unable to read session context from remote: %+v", err))
 	}
+
+	if res.StatusCode == 200 {
+		return body, nil
+	}
+
+	return json.RawMessage{}, errors.WithStack(helper.ErrUnauthorized.WithReasonf("API said: %d -- %+v", res.StatusCode, body))
 }


Makefile
```diff
@@ -79,7 +79,8 @@ install: .bin/packr2
 .PHONY: docker
 docker: .bin/packr2
 		packr2 || (GO111MODULE=on go install github.com/gobuffalo/packr/v2/packr2 && packr2)
-		CGO_ENABLED=0 GO111MODULE=on GOOS=linux GOARCH=amd64 go build
+		CGO_ENABLED=0 GO111MODULE=on GOOS=linux GOARCH=amd64 go build \
```
till, line 82: Hope you don't mind — it adds a version other than `master` to logs when you run it locally.

pipeline/authn/authenticator_bearer_token.go
```diff
@@ -107,6 +107,10 @@ func (a *AuthenticatorBearerToken) Authenticate(r *http.Request, session *Authen
 		return helper.ErrForbidden.WithReasonf("The configured extra_from GJSON path returned an error on JSON output: %s", err.Error()).WithDebugf("GJSON path: %s\nBody: %s\nResult: %s", cf.ExtraFrom, body, extraRaw).WithTrace(err)
 	}
 
+	if len(subject) == 0 {
```
till, line 110: Before, it would go to `upstream` with an empty subject. Which seems not a desired outcome.

pipeline/authn/authenticator_cookie_session.go
```diff
@@ -155,6 +155,10 @@ func forwardRequestToSessionStore(r *http.Request, checkSessionURL string, prese
 
 	// We need to make a COPY of the header, not modify r.Header!
 	for k, v := range r.Header {
+		// remove Accept-Encoding to let the transport handle gzip
+		if k == "Accept-Encoding" {
```
till, line 159: This took us a bit to figure out — it seems that when this is set, it turns off automatic decompressing. So omitting it, lets it do what it needs to do. (also h/t @rwos for finding the associated ticket and helping me fix this)

pipeline/authn/authenticator_cookie_session.go
```diff
@@ -171,13 +175,14 @@ func forwardRequestToSessionStore(r *http.Request, checkSessionURL string, prese
 		return nil, helper.ErrForbidden.WithReason(err.Error()).WithTrace(err)
 	}
 
+	body, err := ioutil.ReadAll(res.Body)
```
till, line 178: I figured, if the body cannot be decoded, that would be a general error and it would be good to bubble up. The changes below it hopefully make reading the flow easier.

Makefile
aeneasr: 👍

pipeline/authn/authenticator_bearer_token.go
aeneasr: Nice!

pipeline/authn/authenticator_bearer_token.go
```diff
@@ -107,6 +107,10 @@ func (a *AuthenticatorBearerToken) Authenticate(r *http.Request, session *Authen
...
+		return helper.ErrForbidden.WithReasonf("Could not decode subject from response.")
```
aeneasr, line 111: NIce! Can you please add a test for this behavior?

Also, this needs a trace and maybe a debug help message :)

```suggestion
		return errors.WithStack(helper.ErrForbidden.WithReasonf("Subject field from remote endpoint is empty."))
```

pipeline/authn/authenticator_cookie_session.go
```diff
@@ -171,13 +175,14 @@ func forwardRequestToSessionStore(r *http.Request, checkSessionURL string, prese
...
-		body, err := ioutil.ReadAll(res.Body)
-		if err != nil {
-			return json.RawMessage{}, errors.WithStack(herodot.ErrInternalServerError.WithReasonf("Unable to fetch cookie session context from remote: %+v", err))
-		}
 		return body, nil
-	} else {
-		return json.RawMessage{}, errors.WithStack(helper.ErrUnauthorized)
 	}
+
+	return json.RawMessage{}, errors.WithStack(helper.ErrUnauthorized.WithReasonf("API said: %d -- %+v", res.StatusCode, body))
```
aeneasr, line 187: Please return a generic error here. Printing out the API response to the caller is dangerous, as it can be used to discover the API responses of the API endpoint by someone who is not allowed to see them usually.

pipeline/authn/authenticator_cookie_session.go
```diff
@@ -171,13 +175,14 @@ func forwardRequestToSessionStore(r *http.Request, checkSessionURL string, prese
...
+	}
+
 	if res.StatusCode == 200 {
-		body, err := ioutil.ReadAll(res.Body)
-		if err != nil {
-			return json.RawMessage{}, errors.WithStack(herodot.ErrInternalServerError.WithReasonf("Unable to fetch cookie session context from remote: %+v", err))
-		}
 		return body, nil
-	} else {
-		return json.RawMessage{}, errors.WithStack(helper.ErrUnauthorized)
```
aeneasr, line 181: Maybe instead you just log the body output?

aeneasr (changes requested):

till:
@aeneasr tried to address everything, let me know what you think.

pipeline/authn/authenticator_cookie_session.go
till: @aeneasr just fyi, I also added a test the header is stripped (and stays stripped). :)

pipeline/authn/authenticator_cookie_session.go
ant1k9: It's a good point to move `ReadAll` from the if statement, because if we won't read the `res.Body` in the else statement, we may have a goroutine leak.

I've come across this problem working with the oathkeeper, but it seems that we also need to defer the `res.Body.Close()` to close the response body explicitly. What do you think about adding a deferred statement with the `Close` method on the line above?

till: @ant1k9 yeah, maybe — @aeneasr thoughts?

till:
@aeneasr we are relying on this to work, is it possible to get this integrated soon? As we are now having issues with websockets (see ory/cloud#76), it's getting messy to maintain this downstream. We'd also like to update of course.

aeneasr:
Sorry for dropping the ball on this - I'm currently on my holidays but will try to find some time to review it :)

pipeline/authn/authenticator_bearer_token.go
aeneasr: Upon review I'm wondering if we really want to do this? An empty subject could have meaning such as "anonymous" user in some environment. Changing this to error will cause a breaking change for those relying on this feature. What was the intention of changing this?

pipeline/authn/authenticator_cookie_session.go
aeneasr: Nice find! I assume this is similar to https://github.com/ory/oathkeeper/pull/952#discussion_r850728149 ?

pipeline/authn/authenticator_cookie_session.go
aeneasr: Yes please defer with body.Close!

pipeline/authn/authenticator_cookie_session.go
```diff
@@ -171,13 +179,16 @@ func forwardRequestToSessionStore(r *http.Request, checkSessionURL string, prese
...
 	}
+
+	logger.Tracef("Remote returned non-200 status code '%d' with body: %s", res.StatusCode, body)
```
aeneasr, line 192: It would be a bit better to use structured logging like `logger.WithField("response_code", res.StatusCode).WithField("body", string(body)).Trace` :)

aeneasr (changes requested):

pipeline/authn/authenticator_cookie_session.go
till: @aeneasr fixed

pipeline/authn/authenticator_cookie_session.go
till: Fixed!

aeneasr:
Hi @till - sorry that I missed your comment. If you resolve the conflicts with master everything should pass and we can take this for another review spin :)

till:
@aeneasr seems like you implemented most of what we had (around preserving headers, settings, etc.) so this PR just adds test coverage for the original issue and fixes a bug where subject is not returned (where it is expected).

aeneasr (dismissed):

PR #952, opened 2022-04-11
feat: pass only essential and configured headers to authenticator

gen1us2k:
BREAKING CHANGE: From now on, the `bearer_token` and `cookie_session` handlers pass only the needed header (`Authorization`, `Cookie`) to the check URL. To pass additional headers, use the `forward_http_headers` configuration key.

Closes https://github.com/ory/oathkeeper/pull/954
Closes https://github.com/ory/cloud/issues/76


gen1us2k:
I tried to use oathkeeper with Ory cloud for the local project. I set up everything and it worked except the header mutator. 
After debugging, I found that the response from Ory Cloud comes with Accept-Encoding gzip header and all requests are gzipped by default.

Another solution to solve that case is to create an HTTP client with the DisableCompression option for [http.Client.Transport](https://go.dev/src/net/http/transport.go#L173)

> // DisableCompression, if true, prevents the Transport from
> // requesting compression with an "Accept-Encoding: gzip"
> // request header when the Request contains no existing
> // Accept-Encoding value. If the Transport requests gzip on
> // its own and gets a gzipped response, it's transparently
> // decoded in the Response.Body. However, if the user
> // explicitly requested gzip it is not automatically
> // uncompressed.

but it doesn't handle deflate encoding.

pipeline/authn/authenticator_cookie_session.go
```diff
@@ -171,8 +173,20 @@ func forwardRequestToSessionStore(r *http.Request, checkSessionURL string, prese
 		return nil, helper.ErrForbidden.WithReason(err.Error()).WithTrace(err)
 	}
 
+	defer res.Body.Close()
+
 	if res.StatusCode == 200 {
-		body, err := ioutil.ReadAll(res.Body)
+		var reader io.Reader
+		switch res.Header.Get("Content-Encoding") {
+		case "gzip":
```
aeneasr, line 181: `http.DefaultTransport` automatically handles gzip decoding if the correct headers are set: https://stackoverflow.com/questions/13130341/reading-gzipped-http-response-in-go

So this should not be required!

aeneasr (changes requested):

pipeline/authn/authenticator_cookie_session.go
gen1us2k: the `http.DefaultTransport` does not use the `DisableCompression` property, and it's disabled by default. Also, the standard HTTP client does not support automatic decompression of `deflate` encoding. 

I can give you a way how to reproduce the issue because I reproduced it by having this setup.

1. Ory proxy proxies traffic from Ory Cloud
2. Oathkeeper uses ory proxy endpoint and cookie_session authenticator
3. Oathkeeper as one upstream to proxy traffic to
I found that with mutator header `X-User` that comes from the cookie_authenticator is always empty and after a couple of hours of debugging I found that's a gzip issue.

I think that I'll write a couple of tests to prove the bug, and maybe I'll find a better solution, but also a couple of users had the same issue in our Slack community

gen1us2k: However, the default client[ handles this by itself ](https://go.dev/src/net/http/transport.go#L2196). I think I'll start with the tests first and try to reproduce it on a clean setup

aeneasr: Ok, that sounds great! We had issues with GZIP in the past in Ory Oathkeeper. I think it depends a bit on where in the pipeline the issue is. I think the proxy had issues with GZIP as it did not decode them automatically - here however I think we do?

Another problem that can happen is that we forward requests with all the headers that might include e.g. `Content-Encoding: gzip` but we actually do not use gzip in the body. So basically an incoming request to oathkeeper is:

```
GET /foo
Content-Encoding: gzip

...gzip-body...
```

then Ory Oathkeeper reads that data and decompresses it. Next, it forwards the request to e.g. the session endpoint with all the HTTP headers:

```
GET /sessions/whoami
Content-Encoding: gzip

...here the body is actually not gzip any more because it was decoded in oathkeeper already...
```

which then causes issues at the receiver.

I think that, in the end, we will need to refactor this so that we only send "wanted" headers to the upstream services instead of sending a catch-all. The same underlying issues is the reason for this: https://github.com/ory/oathkeeper/pull/954

eh-steve: I think the problem is the opposite way around... The behaviour of the Go HTTP client depends on whether the `http.Request` explicitly sets the `Accept-Encoding: gzip` header or not. If no header is set, the default http client automatically adds it before firing the request, and transparently handles decompression of the response for you (stripping out any `Content-Encoding` response headers).

If the request header is set though, it assumes you want to handle the decompression manually, and so returns the response compressed.

This can happen and cause problems if the service you're proxying serves compressed content in response to a request with `Accept-Encoding: gzip` (e.g. an nginx ingress with transparent compression enabled) - Oathkeeper would receive the response which would be decompressed, but because the inbound request had the `Accept-Encoding` header set, the go http client would not transparently decompress it.

The offending lines are below, the blind copy of all inbound request headers onto the outbound request on L161. 
```go
	for k, v := range r.Header {
		req.Header[k] = v
	}
```

We should add a check for the `Accept-Encoding` header here, and not copy it, to avoid this

aeneasr: Right, that can also very well be the root cause! I think the effect and best resolution is to only pass essential and configured headers to the upstream and leave the user the ability to pass additional headers through the config too.

.schema/config.schema.json
```diff
@@ -445,6 +445,14 @@
             "POST"
           ]
         },
+        "proxy_headers": {
```
aeneasr, line 448: Let's call this `forwarded_http_headers` :)

aeneasr (commented):
Nice, this goes in a very good direction! It will also need a breaking change note as well as some documentation in ory/docs :)

.schema/config.schema.json
gen1us2k: Yep. That's a much clearer name for the variable :)

gen1us2k:
@aeneasr I think this is ready for review.

pipeline/authn/authenticator_bearer_token_test.go
```diff
@@ -96,7 +96,7 @@ func TestAuthenticatorBearerToken(t *testing.T) {
 					w.WriteHeader(200)
 					w.Write([]byte(`{"sub": "123"}`))
 				},
-				config:    []byte(`{"preserve_query": true}`),
+				config:    []byte(`{"preserve_query": true, "forward_http_headers": ["Authorization"]}`),
```
Benehiko, line 99: Just a question, i see here we are adding `forward_http_headers` for `Authorization` in many of these tests. Aren't they included automatically now?

pipeline/authn/authenticator_bearer_token.go
```diff
@@ -71,6 +86,14 @@ func (a *AuthenticatorBearerToken) Config(config json.RawMessage) (*Authenticato
 	if len(c.SubjectFrom) == 0 {
 		c.SubjectFrom = "sub"
 	}
+	if len(c.ForwardHTTPHeaders) == 0 {
+		c.ForwardHTTPHeaders = []string{header.Authorization, header.Cookie}
+	}
```
Benehiko, line 91: Just to make sure I'm understanding this correctly, if we have no `forward_http_headers` we would always add `Authorization` and `Cookie`. But shouldn't that be the case for `bearer` and `cookie` authenticators anyway, even if we add more `forward_http_headers`?

referring to this discussion https://github.com/ory/oathkeeper/pull/952#discussion_r868183613

pipeline/authn/authenticator_cookie_session.go
```diff
@@ -75,6 +89,14 @@ func (a *AuthenticatorCookieSession) Config(config json.RawMessage) (*Authentica
 	if len(c.SubjectFrom) == 0 {
 		c.SubjectFrom = "subject"
 	}
+	if len(c.ForwardHTTPHeaders) == 0 {
+		c.ForwardHTTPHeaders = []string{header.Authorization, header.Cookie}
+	}
```
Benehiko, line 94: same question here as with `bearer` authenticator

Benehiko (changes requested):
I think this looks quite good :)
Just a few questions

pipeline/authn/authenticator_bearer_token_test.go
gen1us2k: Yep. I wrote tests first and later I decided to implement some backward compatibility. I can remove it from the tests if you want.

pipeline/authn/authenticator_bearer_token.go
gen1us2k: Yeah. I found the issue here. Right now oathkeeper adds `Authorization` and `Cookie` headers only when we have no `forward_http_headers` configured. However, I think that the oathkeeper should have them in any case because they're essential. 

The issue may occur when someone configures something like
```
forward_http_headers:
   - Accept
   - X-User
```
In that case, the authentication flow breaks because the oathkeeper does not proxies essential headers.

gen1us2k: I think that we can remove `if len(c.ForwardHTTPHeaders) == 0` and append `[]string{header.Authorization, header.Cookie}` anyway. WDYT?

Benehiko: i agree since it is required for `bearer` or `cookie` authenticator to work properly

pipeline/authn/authenticator_bearer_token_test.go
Benehiko: yeah i think it makes sense to remove it. otherwise it will just confuse in future.

We should just add a test though to make sure these headers are always present in these authentictors (explicitly even though the other tests infer it).

gen1us2k: Fixed it :)

pipeline/authn/authenticator_bearer_token.go
gen1us2k: Fixed it too

Benehiko (dismissed):
LGTM :)

pipeline/authn/authenticator_bearer_token.go
```diff
@@ -24,15 +25,29 @@ type AuthenticatorBearerTokenFilter struct {
...
+	PreserveHost          bool                        `json:"preserve_host"`
+	ExtraFrom             string                      `json:"extra_from"`
+	SubjectFrom           string                      `json:"subject_from"`
+	ForwardHTTPHeaders    []string                    `json:"forward_http_headers"`
+	SetHeaders            map[string]string           `json:"additional_headers"`
+	ForceMethod           string                      `json:"force_method"`
+	ForwardHTTPHeadersMap map[string]string           `json:"-"`
+}
+
+func (a *AuthenticatorBearerTokenConfiguration) ToAuthenticatorForwardConfig() *AuthenticatorForwardConfig {
```
aeneasr, line 41: A better approach would be to define an interface which is consumed by `forwardRequestToSessionStore`

```
type AuthenticatorForwardConfigProvider interface {
  GetCheckSessionURL() string
  // ...
}

// Type Guard
var _ AuthenticatorForwardConfigProvider = (*AuthenticatorBearerTokenConfiguration)(nil)
```

or alternatively use an embedded struct:

```
type authenticatorForwardConfig struct {
 	CheckSessionURL    string `json:"..."`
}

type AuthenticatorBearerTokenConfiguration struct {
  authenticatorForwardConfig

  // ...
}

// ...
ody, err := forwardRequestToSessionStore(r, cf. authenticatorForwardConfig)
```

That will save some copy & pasting and potential sources of errors (e.g. forgetting to copy a field) and is in general a cleaner approach to Go code :)

pipeline/authn/authenticator_cookie_session.go
```diff
@@ -49,6 +52,17 @@ func NewAuthenticatorCookieSession(c configuration.Provider) *AuthenticatorCooki
 	}
 }
 
+func (a *AuthenticatorCookieSessionConfiguration) ToAuthenticatorForwardConfig() *AuthenticatorForwardConfig {
+	return &AuthenticatorForwardConfig{
+		CheckSessionURL:    a.CheckSessionURL,
+		PreserveQuery:      a.PreserveQuery,
+		PreservePath:       a.PreservePath,
+		PreserveHost:       a.PreserveHost,
+		ForwardHTTPHeaders: a.ForwardHTTPHeadersMap,
```
aeneasr, line 61: It's confusing that we have `ForwardHTTPHeaders` and `ForwardHTTPHeadersMap`, but depending on where this field is used they are used differently. To help others reading your code, it is good to be explicit about it. As mentioned in a comment below, I don't think we need the map at all.

pipeline/authn/authenticator_cookie_session.go
```diff
@@ -129,55 +151,67 @@ func cookieSessionResponsible(r *http.Request, only []string) bool {
...
 	}
 
-	if res.StatusCode == 200 {
-		body, err := ioutil.ReadAll(res.Body)
-		if err != nil {
-			return json.RawMessage{}, errors.WithStack(herodot.ErrInternalServerError.WithReasonf("Unable to fetch cookie session context from remote: %+v", err))
+	// We need to copy only essential and configurable headers
+	for k, v := range r.Header {
+		if _, ok := cf.ForwardHTTPHeaders[k]; ok {
+			req.Header[k] = v
```
aeneasr, line 205: My recommendation is to not use precomputed map here. You don't really save anything - it requires more memory (to store the map) and more CPU time (because of memory allocation). And it's easy to mess up since the "map" is sort of like a cache. It's totally fine to be explicit and have two loops:

```suggestion
	// We need to copy only essential and configurable headers
	for requested, v := range r.Header {
		for _, allowed := range cf.ForwardHTTPHeaders {
			if requested == allowed {
				req.Header[k] = v
```

pipeline/authn/authenticator_cookie_session_test.go
```diff
@@ -249,6 +250,46 @@ func TestAuthenticatorCookieSession(t *testing.T) {
...
+	t.Run("prepare request should return only configured headers", func(t *testing.T) {
+		testCases := []struct {
+			requestHeaders  []string
+			expectedHeaders []string
+			conf            *AuthenticatorForwardConfig
+		}{
+			{
+				requestHeaders:  []string{header.Authorization, header.AcceptEncoding},
+				expectedHeaders: []string{},
+				conf:            &AuthenticatorForwardConfig{},
```
aeneasr, line 263: Shouldn't this be include our fallbacks? See this line:

https://github.com/ory/oathkeeper/pull/952/files#diff-0879cde6bf6394458fb28990a1d91fb15793e155257301d639d99ab6604a3601R90

aeneasr (changes requested):
Thank you @gen1us2k - this will improve things a lot! I have some feedback regarding the code style, I think the functionality is on top!

Also, pelase add some documentation to ory/docs for this new feature!

@Benehiko please check my review, I hope you find it useful to learn what topics are important to look at during code reviews :)

pipeline/authn/authenticator_cookie_session.go
gen1us2k: Without a map, it's easy to have duplicated keys for "Authorization" and "Cookie" headers for backward compatibility. Maps guarantee that all headers are unique. Using `map[string]struct{}` or `map[string]bool` can save space. On the other hand, the oathkeeper needs to make additional checks for backward compatibility which means more `for` loops at the initialization.

gen1us2k: I can unify variable names to make the code easier to understand.

gen1us2k:
@aeneasr It's ready for another round of review.

pipeline/authn/authenticator_bearer_token_test.go
```diff
@@ -160,7 +160,7 @@ func TestAuthenticatorBearerToken(t *testing.T) {
 					w.WriteHeader(200)
 					w.Write([]byte(`{"sub": "123"}`))
 				},
-				config:    []byte(`{"preserve_path": true, "preserve_query": true, "check_session_url": "http://origin-replaced-in-test/configured/path?q=configured-query"}`),
+				config:    []byte(`{"preserve_path": true, "preserve_query": true, "check_session_url": "http://origin-replaced-in-test/configured/path?q=configured-query", "forward_http_headers": ["X-Forwared-Host"]}`),
```
Benehiko, line 163: is this necessary to add here for the test to pass?

pipeline/authn/authenticator_bearer_token.go
```diff
@@ -31,10 +32,33 @@ type AuthenticatorBearerTokenConfiguration struct {
 	PreserveHost        bool                        `json:"preserve_host"`
 	ExtraFrom           string                      `json:"extra_from"`
 	SubjectFrom         string                      `json:"subject_from"`
+	ForwardHTTPHeaders  []string                    `json:"forward_http_headers"`
 	SetHeaders          map[string]string           `json:"additional_headers"`
 	ForceMethod         string                      `json:"force_method"`
 }
```
Benehiko, line 39: The same here, we need a type guard

```go
var _ AuthenticatorForwardConfig = new(AuthenticatorBearerTokenConfiguration)
```

Benehiko (changes requested):
A lot better :)

Just a few more questions and minor changes

pipeline/authn/authenticator_bearer_token_test.go
Benehiko: i removed it and didn't get any failure

gen1us2k: Yep. You're right. I removed it.

Benehiko (approved):
nice :)

pipeline/authn/authenticator_cookie_session_test.go
```diff
@@ -249,6 +250,56 @@ func TestAuthenticatorCookieSession(t *testing.T) {
...
+				expectedHeaders: []string{},
+				conf:            &AuthenticatorCookieSessionConfiguration{},
+			},
+			{
+				requestHeaders:  []string{header.Authorization, header.AcceptEncoding},
+				expectedHeaders: []string{header.AcceptEncoding},
+				conf: &AuthenticatorCookieSessionConfiguration{
+					// This value is coming from the configuration and may use incorrect casing.
+					ForwardHTTPHeaders: []string{
+						"acCept-enCodinG",
```
aeneasr, line 271: I added this test to show that the header config needs to match the canonical values exactly or it won't work. Can you please fix that?

x/header/header.go
```diff
@@ -0,0 +1,75 @@
+package header
+
+import "net/textproto"
+
+const (
```
aeneasr, line 5: Do we need all of these values? I think we're only using one or two of those constants. When in doubt, leave it out :)

aeneasr (changes requested):
Thank you, we're getting there :) Can you please take a look at the comments? Additionally, can you please add two tests:

1. A test for bearer token where a custom header is forwarded
1. A test for session cookie where a custom header is forwarded

Regarding the CI format failure, @kevgo will take a look

x/header/header.go
gen1us2k: I think it would be great to have those constants and I added the most common headers. I can go with a couple of them.

pipeline/authn/authenticator_cookie_session_test.go
gen1us2k: Fixed it

aeneasr (approved):
Great job :)

PR #954, opened 2022-04-13
fix: don't pass websocket headers on to session store

rwos:
When Oathkeeper is used to secure the initial connection request to a websocket endpoint, that doesn't mean we also want to open a websocket connection to the session store. With this commit, the "Connection: Upgrade" and related headers are removed now, so the session store receives a "normal" HTTP request.

```
BREAKING CHANGES: With this patch, the "Connection", "Upgrade", "Sec-Websocket-Key", and "Sec-Websocket-Version" HTTP Headers are not relayed to the session store anymore.
```

## Related issue(s)

https://github.com/ory/cloud/issues/76

## Checklist

- [x] ~~I have referenced an issue containing the design document if my change
      introduces a new feature.~~ not a new feature
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/pipeline/authn/authenticator_cookie_session.go
+++ b/pipeline/authn/authenticator_cookie_session.go
@@ -155,6 +155,10 @@ func forwardRequestToSessionStore(r *http.Request, checkSessionURL string, prese
 
 	// We need to make a COPY of the header, not modify r.Header!
 	for k, v := range r.Header {
+		// remove websocket-related headers to not confuse reverse proxies
+		if k == "Upgrade" || k == "Connection" || k == "Sec-Websocket-Key" || k == "Sec-Websocket-Version" {
+			continue
+		}
 		req.Header[k] = v
 	}
 
--- a/pipeline/authn/authenticator_bearer_token_test.go
+++ b/pipeline/authn/authenticator_bearer_token_test.go
@@ -223,6 +223,34 @@ func TestAuthenticatorBearerToken(t *testing.T) {
 				},
 			},
 			{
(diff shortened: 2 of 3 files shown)


pipeline/authn/authenticator_cookie_session.go
```diff
@@ -155,6 +155,10 @@ func forwardRequestToSessionStore(r *http.Request, checkSessionURL string, prese
 
 	// We need to make a COPY of the header, not modify r.Header!
 	for k, v := range r.Header {
+		// remove websocket-related headers to not confuse reverse proxies
+		if k == "Upgrade" || k == "Connection" || k == "Sec-Websocket-Key" || k == "Sec-Websocket-Version" {
```
aeneasr, line 159: One problem I realized just now is that the header key might not be canonicalized, meaning that it could also start with a lower case so e.g. `upgrade` or `uPGRADE` depending on how buggy the client is. If you change this to:

```
switch strings.ToLower(k) {
case "upgrade":
  ...
}
```

this should no longer be an issue!

rwos: Go should canonicalize these at request parse time, if I read this right: https://cs.opensource.google/go/go/+/refs/tags/go1.18.1:src/net/http/request.go;l=138-160

rwos: 👍 agreed, I'll try to figure out a test that actually tests this (I think `Header.Set` also canonicalizes 🤪)

PR #958, opened 2022-04-22
feat: allow fallback when using JWT

jnodorp-jaconi:
We try to use the `jwt` authenticator alongside the `bearer_token` authenticator. Our application receives requests with different token type, both in the `Authorization` header. The `bearer_token` tokens are not JWTs.

Currently this is not possible, as the JWT authenticator declares itself responsible as long as _any_ token is provided at the correct location.

With the proposed change, the JWT authenticator will only declare itself responsible if the token at least looks like a JWT (`<header>.<payload>.<signature>`). That way, a subsequent authenticator can attempt authentication with the same token.

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/pipeline/authn/authenticator_jwt.go
+++ b/pipeline/authn/authenticator_jwt.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"strings"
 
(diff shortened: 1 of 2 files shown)


aeneasr (approved):
Great idea!

PR #967, opened 2022-05-18
feat: Add support for header in rule matching

m-guesnon-pvotal:
## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.

### DIFF
--- a/rule/rule.go
+++ b/rule/rule.go
@@ -23,6 +23,7 @@ package rule
 import (
 	"encoding/json"
 	"fmt"
+	"net/http"
 	"net/url"
 	"strings"
 
@@ -35,7 +36,7 @@ type Match struct {
 	// An array of HTTP methods (e.g. GET, POST, PUT, DELETE, ...). When ORY Oathkeeper searches for rules
 	// to decide what to do with an incoming request to the proxy server, it compares the HTTP method of the incoming
 	// request with the HTTP methods of each rules. If a match is found, the rule is considered a partial match.
-	// If the matchesUrl field is satisfied as well, the rule is considered a full match.
+	// If the matchesUrl and matchesHeaders fields are satisfied as well, the rule is considered a full match.
 	Methods []string `json:"methods"`
 
 	// This field represents the URL pattern this rule matches. When ORY Oathkeeper searches for rules
@@ -50,6 +51,14 @@ type Match struct {
 	// The following regexp example matches all paths of the domain `mydomain.com`: `https://mydomain.com/<.*>`.
 	// The glob equivalent of the above regexp example is `https://mydomain.com/<*>`.
 	URL string `json:"url"`
+
+	// A map of HTTP headers. When ORY Oathkeeper searches for rules
+	// to decide what to do with an incoming request to the proxy server, it compares the HTTP headers of the incoming
+	// request with the HTTP headers of each rules. If a match is found, the rule is considered a partial match.
+	// For headers with values in array format (e.g. User-Agent headers), the rule header value must match at least one
+	// of the request header values.
+	// If the matchesUrl and matchesMethods fields are satisfied as well, the rule is considered a full match.
+	Headers map[string]string `json:"headers"`
 }
 
 type Handler struct {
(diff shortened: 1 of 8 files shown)


m-guesnon-pvotal:
Documentation pull request can be [found here](https://github.com/ory/docs/pull/819)

m-guesnon-pvotal:
Formatting with npm run format to please the CI modified all the *.md file. This can be reverted if needed.

aeneasr (changes requested):
Could you please revert all changes not related to this PR? :)

m-guesnon-pvotal:
Done. However, this brought back the linter fail for CI. (I suspect current master could not pass the CI)

a-manraj-pvotal:
@aeneasr any news for linter and the content of the PR itself ?

a-manraj-pvotal:
Outside linting @aeneasr any changes required or discussion to alter ?

rule/rule.go
```diff
@@ -50,6 +51,14 @@ type Match struct {
 	// The following regexp example matches all paths of the domain `mydomain.com`: `https://mydomain.com/<.*>`.
 	// The glob equivalent of the above regexp example is `https://mydomain.com/<*>`.
 	URL string `json:"url"`
+
+	// A map of HTTP headers. When ORY Oathkeeper searches for rules
+	// to decide what to do with an incoming request to the proxy server, it compares the HTTP headers of the incoming
+	// request with the HTTP headers of each rules. If a match is found, the rule is considered a partial match.
+	// For headers with values in array format (e.g. User-Agent headers), the rule header value must match at least one
+	// of the request header values.
+	// If the matchesUrl and matchesMethods fields are satisfied as well, the rule is considered a full match.
```
aeneasr, line 60: Wha is the difference between a partial and a full match?

rule/rule.go
```diff
@@ -50,6 +51,14 @@ type Match struct {
...
+	Headers map[string]string `json:"headers"`
```
aeneasr, line 61: As far as I understand, this is an `AND` match. So if all requirements are met from the rule, it will apply. Is that desired? Are there use cases for `OR`? And if so, how do we handle those? We can also only support `AND` for now but will nede to support `OR` most likely at some point, without breaking the config.

aeneasr (changes requested):
Sorry for the super late review, my backlog is simply too big at the moment. I have a couple of questions for this PR and would be thankful if you could answer those

rule/rule.go
m-guesnon-pvotal: At this point, the implementation only supports `AND`.
You have to match all headers specified in the rule. IMHO, an `OR` implementation would be overcomplicating things. It should just be a different rule. Headers are seen just like URLs are (although, it is true that URLs support wildcards).

rule/rule.go
m-guesnon-pvotal: This is wording have taken from the comments on other fields.
My interpretation is that a partial match means you satisfied to one aspect of the rule. The incoming request has to be a (partial) match to all aspects of this rule in order to become a full match.

rule/rule.go
a-manraj-pvotal: I think that using using `OR` is definitely a breaking change as it would potentially override normal oathkeeper rule matching and documentation with an optional routing header. Therefore i would recommend supporting only `AND` for now as an advanced configuration parameter.

aeneasr: Ok, that makes sense!

aeneasr:
If you rebase on master, I think that the CI should pass!

aeneasr:
I'm unfortunately not allowed to do so :(

m-guesnon-pvotal:
@vinckr Done. Sorry, I missed @aeneasr message and never followed up.

timdrijvers:
This would be a great enhancement of Oathkeeper, is there any chance this will be picked up (or supported in a different way)?

vinckr:
Hello @timdrijvers 
definitely! 
As soon as the team can make some time to work on it - we can't commit to any timelines on here though, right now there is a ton of work in the backlog for customers of Ory Network/Ory Enterprise License.  
If you have an Ory support agreement or Enterprise License feel free to reach out to your account manager to prioritize this feature.

PR #978, opened 2022-06-27
ci: improve formatting checks

kevgo:
CI and CLI now always use the same version of Prettier and its shared config: the one in `package.json`. CI now runs `make format` instead of just Prettier, i.e. it now also runs `gofmt` and the `goimports` checker. To verify formatting, CI now formats the code and checks for uncommitted changes afterwards. This provides two benefits: 

1. CI can show a diff of the changes needed to make it pass. 
   This is the information needed to fix formatting problems manually.
   This allows us to not commit fixes on CI when submitting tiny documentation updates via the GitHub UI.
2. We can get rid of the `npm format:check` script, which was really only useful on CI.
   This script was highly redundant with `npm format`.
   Developers now simply run `make format` when they are unsure whether their code is formatted.

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.


kevgo:
format

kevgo:
format

kevgo:
format

.github/workflows/ci.yml
```diff
@@ -25,25 +25,32 @@ jobs:
...
+          path: node_modules
+          key: ${{ runner.os }}-node-${{ hashFiles('package-lock.json') }}
       - uses: actions/setup-go@v2
         with:
           go-version: ^1.17
+      - run: npm ci
+        if: steps.cache-node.outputs.cache-hit != 'true'
+      - run: make format
+      - name: Indicate formatting problems
+        run: git diff HEAD --exit-code
```
kevgo, line 50: I was considering to extract this into a shared action in https://github.com/ory/ci but I think being explicit is better here.

kevgo:
Closing and reopening to trigger the new format check to run.

kevgo:
@aeneasr okay, I have fixed everything. Initially I followed the old setup but the `CI` status check creates uncommitted files on CI which leads to false positives in the format check. Long story short, the format check is now in a dedicated status check called `Format` and everything works.

While closing out this PR, can you please also remove the old required status checks that no longer exist (see attached screenshot)? Also, maybe make some of the existing status checks required again?

Screenshot from 2022-07-14 18-18-24

aeneasr:
thanks!

PR #982, opened 2022-07-04
feat: introduce auth scheme and jumping to next authentication

Marlinc:
Some (legacy) application don't use `Bearer` as auth scheme in the `Authorization` header, in the current implementation this means Oathkeeper cannot be used to handle these legacy applications. The bearer token (and cookie) authenticator currently also don't support handling multiple authentication configurations (for example an access token in a header but also in a query parameter).

This change adds an optional `auth_scheme` option to the bearer token authenticator that allows changing the scheme it accepts, it also supports setting an empty scheme for applications that provide a token directly in a header. For compatibility it defaults to `Bearer` when the header is set to `Authorization`. The second change is that the session store endpoint can now return a `HTTP 406 Not Acceptable` response to tell Oathkeeper to jump to the next authentication rule.

This changes were previously discussed on [Slack](https://ory-community.slack.com/archives/C012USDT5QQ/p1645536113144619).

I will add some documentation too about these new features.

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/.schema/config.schema.json
+++ b/.schema/config.schema.json
@@ -456,6 +456,11 @@
                   "title": "Header",
                   "type": "string",
                   "description": "The header (case insensitive) that must contain a token for request authentication.\n It can't be set along with query_parameter or cookie."
+                },
+                "auth_scheme": {
+                  "title": "Auth scheme",
+                  "type": "string",
+                  "description": "The auth scheme to accept in case the header is set to Authorization, this is by default set to Bearer."
                 }
               }
             },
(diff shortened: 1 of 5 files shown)


aeneasr (commented):
Thank you @Marlinc and sorry. for the late review! I'm wondering if it would make more sense to use a generic authenticator in such a case? The bearer token authenticator is supposed to solve https://datatracker.ietf.org/doc/html/rfc6750 which is expecting the `Bearer` to be in the header. If that's not the case, wouldn't it be a different type of authentication (same mechanism maybe, but a different HTTP auth mechanism).

We could add another authenticator that handles this (e.g. a  more generic one)

aeneasr:
This would potentially be solved by: https://github.com/ory/oathkeeper/pull/949

PR #997, opened 2022-08-30
feat: add parameter to hide request paths in metrics

adamwalach:
This PR adds `hide_request_paths` parameter to remove request paths from metrics. In some scenarios, exposing paths has very big impact on metrics cardinality. 

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.

### DIFF
--- a/metrics/middleware.go
+++ b/metrics/middleware.go
@@ -76,7 +76,7 @@ func (m *Middleware) CollapsePaths(flag bool) *Middleware {
 }
 
 // HidePaths if set to true, forces the value of the "request" label
-// of the prometheus request metrics to be set to a constant "hidden" value.
+// of the prometheus request metrics to be set to an empty value.
 // eg. (when set to true):
 //   - /decisions/service/my-service -> /hidden
 //   - /decisions -> /hidden
--- a/metrics/middleware.go
+++ b/metrics/middleware.go
@@ -38,6 +38,7 @@ type Middleware struct {
 	mutex         sync.RWMutex
 	silencePaths  map[string]bool
 	collapsePaths bool
+	hidePaths     bool
 }
 
 // NewMiddleware returns a new *Middleware, yay!
@@ -48,6 +49,7 @@ func NewMiddleware(prom *PrometheusRepository, name string) *Middleware {
 		clock:         &realClock{},
 		silencePaths:  map[string]bool{},
 		collapsePaths: true,
+		hidePaths:     true,
 	}
 }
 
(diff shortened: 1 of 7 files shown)


aeneasr (commented):
Thank you! I would actually also be fine to not make this configurable, or alternatively to use a configurable list which would allow us to specify which paths we care about. WDYT?

adamwalach:
I don't want to introduce any regression, someone might be using the `request` label in their infrastructure. That's why the new parameter is false by default and metrics are unchanged unless someone decides otherwise. 

We already have the `collapsePaths` option enabled by default - it leaves just the root part of the URL path.
That means we could implement the list of allowed paths and in combination with `collapsePaths` we get only a handful of values which will keep cardinality of this metric on low level. 
BUT :) I think that this kind of filtering can (or should?) be implemented on WAF level, so any request that tries to access an undefined path is blocked before it reaches Oathkeeper. If we get only filtered paths, then the metric cardinality will stay low. 

In summary, I think that this PR can provide a quick solution to the metric size problem without introducing too much potentially unnecessary changes in Oathkeeper. 
But that's just my opinion :)

aeneasr:
Thank you! 

> I don't want to introduce any regression, someone might be using the request label in their infrastructure. That's why the new parameter is false by default and metrics are unchanged unless someone decides otherwise.

From my side that would be ok, but let's make the transition smooth and remove the option at some point :)

> BUT :) I think that this kind of filtering can (or should?) be implemented on WAF level, so any request that tries to access an undefined path is blocked before it reaches Oathkeeper. If we get only filtered paths, then the metric cardinality will stay low.

The problem is that it is possible to create an infinite number of valid paths even if we filter out bad paths. That is because we have request parameters and therefore you can construct any path: `/identities/abcdef...` for example. That problem of high cardinality was already reported by users with a high volume where you have potentially millions of clients, identities, etc.

Therfore I think that if we want discoverable paths, we need to collapse them in oathkeeper or another type of filter in prometheus (or wherever).

adamwalach:
> Thank you!
> 
> > I don't want to introduce any regression, someone might be using the request label in their infrastructure. That's why the new parameter is false by default and metrics are unchanged unless someone decides otherwise.
> 
> From my side that would be ok, but let's make the transition smooth and remove the option at some point :)
Ok, I'll create a follow-up issue about this

> > BUT :) I think that this kind of filtering can (or should?) be implemented on WAF level, so any request that tries to access an undefined path is blocked before it reaches Oathkeeper. If we get only filtered paths, then the metric cardinality will stay low.
> 
> The problem is that it is possible to create an infinite number of valid paths even if we filter out bad paths. That is because we have request parameters and therefore you can construct any path: `/identities/abcdef...` for example. That problem of high cardinality was already reported by users with a high volume where you have potentially millions of clients, identities, etc.
> 
> Therfore I think that if we want discoverable paths, we need to collapse them in oathkeeper or another type of filter in prometheus (or wherever).

We already collapse them by default: https://github.com/ory/oathkeeper/blob/master/metrics/middleware.go#L67

metrics/middleware.go
```diff
@@ -73,6 +75,18 @@ func (m *Middleware) CollapsePaths(flag bool) *Middleware {
 	return m
 }
 
+// HidePaths if set to true, forces the value of the "request" label
+// of the prometheus request metrics to be set to an empty value.
+// eg. (when set to true):
+//   - /decisions/service/my-service -> /hidden
+//   - /decisions -> /hidden
```
aeneasr, line 82: Shouldn't this be

```suggestion
//   - /decisions/service/my-service -> /
//   - /decisions -> /
```

?

adamwalach: Right, this is a leftover from my tests. At the beginning, I used a "/hidden" value for a path to validate if I see it correctly in prometheus. I will remove the example, I think the explanation mentioning "empty value" should be sufficient.

aeneasr (approved):
LGTM

PR #1004, opened 2022-09-06
feat: add gRPC middleware

hperl:
This PR adds a gRPC middleware to Oathkeeper to use in other services.

This is the gRPC middleware commit cherry-picked on top of the dependency bump.


rule/rule.go
```diff
@@ -143,25 +165,57 @@ func (r *Rule) UnmarshalJSON(raw []byte) error {
...
 
-	if err := errors.WithStack(json.Unmarshal(transformed, &rr)); err != nil {
-		return err
+	if err := json.Unmarshal(transformed, &rr); err != nil {
+		return errors.WithStack(err)
 	}
 
-	*r = rr
+	// polymorphic decoding of the match based on keys
+	if rr.RawMatch != nil && len(rr.RawMatch) > 0 {
```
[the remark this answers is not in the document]
hperl: agreed!

middleware/grpc_middleware_test.go
```diff
@@ -0,0 +1,310 @@
...
+
+					_, err := client.EmptyCall(ctx, &grpcTesting.Empty{})
+					tc.assert(t, err)
+
+					_, err = client.UnaryCall(ctx, &grpcTesting.SimpleRequest{})
+					assertErrDenied(t, err)
+
+					stream, _ := client.StreamingOutputCall(ctx, &grpcTesting.StreamingOutputCallRequest{})
+					_, err = stream.Recv()
+					assertErrDenied(t, err)
```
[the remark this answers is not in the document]
hperl: Right now this is out of scope of what the middleware can do (it will always deny streaming requests).

We don't need them right now so I didn't think too hard about whether the semantics translate nicely or if we need another field (now that we have the special `MatchGRPC` type).

WDYT? I would like to defer streaming to later when we understand the constrains better.

middleware/definitions.go
```diff
@@ -0,0 +1,77 @@
...
+	Option func(*options)
+)
+
+// WithLogger uses the specified logger in the middleware.
+func WithLogger(l *logrusx.Logger) Option {
+	return func(o *options) { o.logger = l }
+}
+
+// WithConfig sets the path to the config file to use for the middleware.
+func WithConfig(configFile string) Option {
```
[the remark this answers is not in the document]
hperl: agreed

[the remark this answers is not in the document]
hperl: agreed

middleware/grpc_middleware_test.go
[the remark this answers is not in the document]
zepatrik: OK makes sense, I just thought the streaming interceptor is already there and works? But it's fine by me to not have support for now.

middleware/grpc_middleware_test.go
```diff
@@ -0,0 +1,310 @@
...
+					},
+					Authenticators: []rule.Handler{{Handler: "bearer_token"}},
+					Authorizer:     rule.Handler{Handler: "allow"},
+					Mutators:       []rule.Handler{{Handler: "noop"}},
+				}},
+			},
+			dialOpts: []grpc.DialOption{grpc.WithPerRPCCredentials(
+				testToken("incorrect token"))},
+			assert: assertErrDenied,
+		},
```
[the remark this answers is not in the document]
hperl: agreed

middleware/grpc_middleware_test.go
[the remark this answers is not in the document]
hperl: It's there, but are the semantics the same? For example, if streaming connections are open for a long time, do we want to re-validate the checks at some point?

F.e., we might want to close streaming connections if a user session is closed.

That's what I meant by semantics are still unclear for me.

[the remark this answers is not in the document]
zepatrik: Yes, what you implemented right now does a check on every message submitted through the stream, which might get very slow (especially with `remote_json` or similar rules).

zepatrik (dismissed):
Awesome, looks good to me now :tada:

aeneasr:
Can you please add documentation for this feature in the oathkeeper docs?

hperl:
@aeneasr docs PR is in https://github.com/ory/docs/pull/985

middleware/definitions.go
```diff
@@ -0,0 +1,77 @@
...
+	o := options{
+		logger:     logrusx.New("Ory Oathkeeper Middleware", x.Version),
+		configFile: "oathkeeper.yaml",
+	}
+	for _, opt := range opts {
+		opt(&o)
+	}
+
+	c, err := configuration.NewKoanfProvider(
+		ctx, nil, o.logger,
```
aeneasr, line 60: Is the nil flag ok here? I'm not sure what impact it has

middleware/definitions.go
```diff
@@ -0,0 +1,77 @@
...
+		configx.WithContext(ctx),
+		configx.WithLogger(o.logger),
+		configx.WithConfigFiles(o.configFile),
+	)
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+
+	r := driver.NewRegistry(c).WithLogger(o.logger).WithBuildInfo(x.Version, x.Commit, x.Date)
+	if o.registryAddr != nil {
```
aeneasr, line 70: I couldn't really find a case where this would be true, shall we remove it?

rule/rule.go
```diff
@@ -143,44 +165,78 @@ func (r *Rule) UnmarshalJSON(raw []byte) error {
...
 
 	transformed, err := migrateRuleJSON(raw)
 	if err != nil {
-		return err
+		return errors.WithStack(err)
 	}
 
-	if err := errors.WithStack(json.Unmarshal(transformed, &rr)); err != nil {
-		return err
+	if err := json.Unmarshal(transformed, &rr); err != nil {
```
aeneasr, line 183: I couldn't find a dedicated test for this new functionality, did I miss it?

rule/matcher_test.go
```diff
@@ -105,7 +105,7 @@ func TestMatcher(t *testing.T) {
 	}
 
 	var testMatcher = func(t *testing.T, matcher Matcher, method string, url string, expectErr bool, expect *Rule) {
-		r, err := matcher.Match(context.Background(), method, mustParseURL(t, url))
+		r, err := matcher.Match(context.Background(), method, mustParseURL(t, url), ProtocolHTTP)
```
aeneasr, line 108: I think we should test GRPC match also

aeneasr (changes requested):
Thank you, this looks good! I like the grpc tests :) I think we need a few more tests in the matcher strategy, and we also need to document the new GRPC matcher configuration in the docs :)

middleware/definitions.go
hperl: The nil is the flags, so if used as a middleware Oathkeeper should not consume any flags, i.e., the `--config` flag.

This reminds me to check that it **also** does not consume environment variables. IMHO the only way to configure the middleware should be through the config file, so that no options clash between Oathkeeper and the main application.

middleware/definitions.go
hperl: It is actually nil by default and only set in tests to get a reference to the registry used by the middleware.

Standard use of the middleware will not have any need for direct registry access.

rule/rule.go
hperl: You're right, that was not covered. I added dedicated Unmarshal tests.

middleware/definitions.go
hperl: OK, now I also isolated the ENV. I think it's fine to completely disable env loading (as it is now), but we could also have env vars under a prefix `OATHKEEPER_`, if we really want env support in the middleware. WDYT?

aeneasr: Ok, makes sense for now

aeneasr (approved):

PR #1013, opened 2022-09-15
Force MinVersion to use TLS in version 1.3 to improve in security

CoolerVoid:
So by default, that resource uses TLS version 1.0, which is not secure. The recommendation by OWASP is to use version 1.3. 

Please look at the following cheatsheetseries.owasp.org/cheatsheets/Transport_Layer_Protection_Cheat_Sheet.html

#security


aeneasr (changes requested):
As far as I know, graceful sets this default automatically for us, so this change should not have an effect :)

PR #1016, opened 2022-09-19
fix: init registry in middleware

hperl:
This PR fixes that the registry needs to be `Init()`ed.

## Related issue(s)

blocks https://github.com/ory-corp/cloud/pull/3154

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/credentials/fetcher_default.go
+++ b/credentials/fetcher_default.go
@@ -66,23 +66,31 @@ type FetcherDefault struct {
 	mux         *blob.URLMux
 }
 
+type FetcherOption func(f *FetcherDefault)
+
+func WithURLMux(mux *blob.URLMux) FetcherOption {
+	return func(f *FetcherDefault) { f.mux = mux }
+}
+
 // NewFetcherDefault returns a new JWKS Fetcher with:
 //
 //   - cancelAfter: If reached, the fetcher will stop waiting for responses and return an error.
 //   - waitForResponse: While the fetcher might stop waiting for responses, we will give the server more time to respond
 //     and add the keys to the registry unless waitForResponse is reached in which case we'll terminate the request.
-func NewFetcherDefault(l *logrusx.Logger, cancelAfter time.Duration, ttl time.Duration) *FetcherDefault {
-	return &FetcherDefault{
+func NewFetcherDefault(l *logrusx.Logger, cancelAfter time.Duration, ttl time.Duration, opts ...FetcherOption) *FetcherDefault {
+	f := &FetcherDefault{
 		cancelAfter: cancelAfter,
 		l:           l,
 		ttl:         ttl,
 		keys:        make(map[string]jose.JSONWebKeySet),
 		fetchedAt:   make(map[string]time.Time),
-		client: httpx.NewResilientClient(
-			httpx.ResilientClientWithConnectionTimeout(15 * time.Second),
-		).StandardClient(),
-		mux: cloudstorage.NewURLMux(),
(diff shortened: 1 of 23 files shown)


middleware/grpc_middleware_test.go
```diff
@@ -378,3 +376,57 @@ func TestMiddleware_EnvironmentIsolation(t *testing.T) {
...
+		middleware.WithConfigFile(configFile),
+		middleware.WithRegistry(regPtr),
+	)
+	require.NoError(t, err)
+	reg := *regPtr
+
+	time.Sleep(100 * time.Millisecond)
+
+	_, err = reg.RuleRepository().Get(ctx, "some-rule-id")
+	require.NoError(t, err)
```
zepatrik, line 431: Maybe add a small assertion that verifies the content was loaded?

middleware/grpc_middleware_test.go
```diff
@@ -177,8 +178,8 @@ mutators:
 			rules: map[configuration.MatchingStrategy][]rule.Rule{
 				configuration.Regexp: {{
 					Match: &rule.MatchGRPC{
-						Authority:  "<[0-9a-zA-Z]+>.apis.ory.sh",
-						FullMethod: "grpc.testing.TestService/Empty<[a-zA-Z]+>",
+						Authority:  "<.*>",
+						FullMethod: "<.*>Empty<.*>",
```
zepatrik, line 182: Why did this change?

zepatrik (changes requested):

middleware/grpc_middleware_test.go
hperl: This is closer to the rules I wrote in Cloud, so I tested for it ;)

middleware/grpc_middleware_test.go
hperl: It already verifies it through the id lookup (which failed before), but I can add some more content tests.

Benehiko (commented):
I think this is quite good, I just have some minor comments

pipeline/authz/remote_json_test.go
```diff
@@ -249,7 +249,9 @@ func TestAuthorizerRemoteJSONValidate(t *testing.T) {
 		},
 	}
 	for _, tt := range tests {
+		tt := tt
```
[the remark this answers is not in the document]
hperl: In Go, the loop variable (the `x` in `for _, x := range foo`) always has the same address, just different values per iteration. If we don't capture the loop variable here, the runs will start in parallel and during execution read the value of that loop variable, which might have already changed.

(See also https://eleni.blog/2019/05/11/parallel-test-execution-in-go/ and https://eli.thegreenplace.net/2019/go-internals-capturing-loop-variables-in-closures/)

credentials/fetcher_default.go
```diff
@@ -200,6 +208,7 @@ func (s *FetcherDefault) resolveAll(done chan struct{}, errs chan error, locatio
 	var wg sync.WaitGroup
 
 	for _, l := range locations {
+		l := l
```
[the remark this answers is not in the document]
hperl: I again think it was a bug to leave the loop capture out here, because we use that to start a go routine.

In short, we need to capture the loop variable if we are using something concurrent with it.

pipeline/authz/remote_json_test.go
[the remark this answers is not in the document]
Benehiko: okay nice, i thought it was something like that. Just making sure :)

Benehiko (approved):

zepatrik (approved):

Issue #988, opened 2022-08-02
Hydrator mutator isn't compatible with gzipped responses.


juan131:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/oathkeeper/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/oathkeeper/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe the bug

As it's described in [the documentation](https://www.ory.sh/docs/oathkeeper/pipeline/mutator#hydrator), the hydrator mutator works by sending a POST request containing a JSON representation of AuthenticationSession to the URL you configure (it's also possible to customize the authentication headers if needed). However, the headers to use on that POST request can't be customized.

It seems that Oathkeeper is accepting gzip-encoding but then it's unable to handle it.

### Reproducing the bug

1. I have an Golang API that uses chi's as router and uses the "[Compress](https://pkg.go.dev/github.com/go-chi/chi/middleware#Compress)" middleware (see https://github.com/go-chi/chi#core-middlewares), this middleware compress the HTTP responses using gzip when the requester adds the header `Accept-Encoding`. This API exposes an endpoint that accepts POST requests including a AuthenticationSession (as described in the docs).
2. Run Oathkeeper configuring a Hydrator mutator that uses the API mentioned above.
3. Test the Hydrator mutator by accessing a resources protected by Oathkeeper.
4. Oathkeeper is sending the POST accepting gzipped encoding, but then it's unable to decode the response (see relevant logs).

### Relevant log output

```shell
{
    "error": {
      "message": "invalid character '\\x1f' looking for beginning of value"
    },
    (...)
    "level": "warning",
    "msg": "The mutation handler encountered an error",
    "mutation_handler": "hydrator",
    "reason_id": "mutation_handler_error",
    (...)
}
```

### Relevant configuration

```yml
serve:
  mutators:
    hydrator:
      enabled: true
      config:
        api:
[...]

PR #1018, opened 2022-09-22
Add support for compressed hydrator #988

mvanderlee:
Oathkeeper doesn't restrict the encoding it accepts when making the hydrator HTTP call. If the hydrator is configured with compression it will try to parse the compressed string directly as JSON and fail. 

This PR restricts the accepted encoding types to gzip and deflate, and add the ability to handle those encodings.
I experimented with brotli as well, but that would require CGO, which seemed too big of a change.

## Related issue(s)

#988

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

## Further Comments

I've already deployed this fix in our production environment and it work. 
Don't think we need to update any docs as this fixes a low level bug.

### DIFF
--- a/pipeline/mutate/mutator_hydrator.go
+++ b/pipeline/mutate/mutator_hydrator.go
@@ -27,6 +27,7 @@ import (
 	"crypto/md5"
 	"encoding/json"
 	"fmt"
+	"io"
 	"net/http"
 	"net/url"
 	"time"
--- a/pipeline/mutate/mutator_hydrator.go
+++ b/pipeline/mutate/mutator_hydrator.go
@@ -22,6 +22,8 @@ package mutate
 
 import (
 	"bytes"
+	"compress/gzip"
+	"compress/zlib"
 	"crypto/md5"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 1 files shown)


pipeline/mutate/mutator_hydrator.go
```diff
@@ -215,8 +221,21 @@ func (a *MutatorHydrator) Mutate(r *http.Request, session *authn.AuthenticationS
 		return errors.New(ErrNon200ResponseFromAPI)
 	}
 
+	// Handle compressed data
```
aeneasr, line 224: The go stdlib HTTP client should do that automatically, does it not?

mvanderlee: From the stdlib HTTP [docs](https://pkg.go.dev/net/http#Transport):

> // If the Transport requests gzip on
	// its own and gets a gzipped response, it's transparently
	// decoded in the Response.Body. However, if the user
	// explicitly requested gzip it is not automatically
	// uncompressed.

So because we set the Accept-Encoding header, we must decompress ourselves. 
Based on what I and Juan131 observed, the transport was not decoding the response prior to that either. 

It's been over 2 years since I touched Go and we've been running this code in production all this time.
However, I have not yet tested again with the latest version of Go.

PR #1034, opened 2022-11-15
feat: added allow_if authorizer

akselleirv:
## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

Will update these as soon I'm close to an approval.

### DIFF
--- a/.schemas/authorizers.allow_if.schema.json
+++ b/.schemas/authorizers.allow_if.schema.json
@@ -1,0 +1,30 @@
+{
+  "$id": "https://raw.githubusercontent.com/ory/oathkeeper/master/.schemas/authorizers.allow_if.schema.json",
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "object",
+  "title": "Allow If Authorizer Configuration",
+  "description": "This section is optional when the authorizer is disabled.",
+  "properties": {
+    "key": {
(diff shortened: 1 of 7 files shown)


aeneasr:
Thank you for the PR! Could you please summarize what this feature does and how it works? Thank you! :)

akselleirv:
> Thank you for the PR! Could you please summarize what this feature does and how it works? Thank you! :)

Hello @aeneasr,

One use-case I have is that I want to check if the JWT has the expected subject. For example, one of the rules I have only permits a user which has the `sub` field set to some value.

An example config would look like:

```
- id: check-sub
  upstream:
    url: https://example.com
  match:
    url: https://example.com
    methods:
      - GET
      - POST
  authenticators:
    - handler: jwt
      config:
        allowed_algorithms:
          - RS256
        jwks_urls:
          - https://example.com
        trusted_issuers:
          - https://example.com
        token_from:
          header: X-Id-Token
        target_audience: []
  authorizer:
    - handler: allow_if
      config:
        key: "{{ .Subject }}"
        op: Equals
        value: aeneasr
```

Another use-case would be to check if some header equals a value:

```
  authorizer:
    - handler: allow_if
      config:
        key: "{{ .Header.some-key }}"
        op: Equals
        value: some-value
```

akselleirv:
@aeneasr do you have time to look at this again?

PR #1041, opened 2022-12-13
fix: do not leak sensitive data from gRPC middleware

zepatrik:
Currently, some sensitive headers might be leaked.


aeneasr (commented):
Looking good but test fail :(

zepatrik:
:shrug: cannot reproduce locally

PR #1042, opened 2022-12-17
fix: add caching to koanf config to fix performance regression

### LINKED ISSUE #1033: Performance/CPU usage regression in v0.40.0
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/oathkeeper/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/oathkeeper/blob/master/CONTRIBUTING.md).
- [X] This issue affects my [Ory Network](https://www.ory.sh/) project.
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [X] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe the bug

When upgrading from v0.39.4 to v0.40.0, we experience an increase of CPU load on our Oathkeeper instances.
In our case, our Kubernetes pods were throttled and our response times at least doubled.

There is the experienced increased CPU usage:
image
From roughly 20-30 mcores, to around 200 mcores.

Discussed in the Ory Slack here https://ory-community.slack.com/archives/C01340V8KSM/p1668072872425319

### Reproducing the bug

In version v0.40.0 and in our oathkeeper config we have 3 rules and one mutator.
- `oauth2_introspection`, being the most used
- `cookie_session`
- `bearer_token`
And a mutator that injects an HTTP header.

### Relevant log output

_No response_

### Relevant configuration

_No response_

### Version

v0.40.0

### On which operating system are you observing this issue?

Linux

### In which environment are you deploying?

Kubernetes

### Additional Context

_No response_

David-Wobrock:
Fix https://github.com/ory/oathkeeper/issues/1033 of a performance regression in v0.40.0.
It seems linked to a perpetual config reloading, leading to a lot of JSON marshalling being CPU intensive.
Adding back in-memory caching, as we used to have for Viper configuration.

## Related issue(s)

Fixes #1033

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/driver/configuration/provider_koanf.go
+++ b/driver/configuration/provider_koanf.go
@@ -6,6 +6,7 @@ package configuration
 import (
 	"bytes"
 	"context"
+	"crypto/sha256"
 	"encoding/json"
 	"fmt"
 	"net/url"
@@ -13,6 +14,8 @@ import (
 	"sync"
 	"testing"
 	"time"
+
+	"github.com/dgraph-io/ristretto"
 
 	"github.com/google/uuid"
 	"github.com/knadh/koanf"
@@ -40,11 +43,7 @@ type (
 		l      *logrusx.Logger
 		ctx    context.Context
 
-		enabledMutex sync.RWMutex
-		enabledCache map[uint64]bool
-
-		configMutex sync.RWMutex
-		configCache map[uint64]json.RawMessage
+		configValidationCache *ristretto.Cache
 
 		subscriptions subscriptions
 	}
@@ -60,13 +59,24 @@ type (
 var _ Provider = new(KoanfProvider)
 
 func NewKoanfProvider(ctx context.Context, flags *pflag.FlagSet, l *logrusx.Logger, opts ...configx.OptionModifier) (kp *KoanfProvider, err error) {
+	maxItems := int64(5000)
+	cache, _ := ristretto.NewCache(&ristretto.Config{
+		NumCounters:        maxItems * 10,
+		MaxCost:            maxItems,
+		BufferItems:        64,
+		Metrics:            false,
+		IgnoreInternalCost: true,
+		Cost: func(value interface{}) int64 {
+			return 1
+		},
+	})
+
 	kp = &KoanfProvider{
 		ctx: ctx,
 		l:   l,
 
-		enabledCache:  make(map[uint64]bool),
-		configCache:   make(map[uint64]json.RawMessage),
-		subscriptions: subscriptions{data: make(map[SubscriptionID]callback)},
+		configValidationCache: cache,
+		subscriptions:         subscriptions{data: make(map[SubscriptionID]callback)},
 	}
 	kp.source, err = configx.New(
 		ctx,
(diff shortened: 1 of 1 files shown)


driver/configuration/provider_koanf.go
```diff
@@ -280,7 +288,44 @@ func (v *KoanfProvider) ToScopeStrategy(value string, key string) fosite.ScopeSt
 }
 
 func (v *KoanfProvider) pipelineIsEnabled(prefix, id string) bool {
-	return v.source.Bool(fmt.Sprintf("%s.%s.enabled", prefix, id))
```
aeneasr, line 283: I think this is very fast as it is without caching, leave it as is :)

David-Wobrock: I wasn't sure how long loading the `source.Bool` would take (if it's reloading something or not).

Commit dropped ✔️

driver/configuration/provider_koanf.go
```diff
@@ -64,9 +64,9 @@ func NewKoanfProvider(ctx context.Context, flags *pflag.FlagSet, l *logrusx.Logg
 		ctx: ctx,
 		l:   l,
 
-		enabledCache:  make(map[uint64]bool),
-		configCache:   make(map[uint64]json.RawMessage),
-		subscriptions: subscriptions{data: make(map[SubscriptionID]callback)},
+		configCache:    make(map[uint64]json.RawMessage),
```
aeneasr, line 67: I think it would be better to use Ristretto instead

driver/configuration/provider_koanf.go
```diff
@@ -283,12 +289,46 @@ func (v *KoanfProvider) pipelineIsEnabled(prefix, id string) bool {
...
+	ts := v.configChangeAt.UnixNano()
+	b := make([]byte, 8)
+	binary.LittleEndian.PutUint64(b, uint64(ts))
+
+	slices := [][]byte{
+		[]byte(prefix),
+		[]byte(id),
+		[]byte(override),
+		b,
+	}
```
aeneasr, line 302: This cache key will break hot reloading, because the cache key does not respect the config values.

driver/configuration/provider_koanf.go
```diff
@@ -306,7 +346,11 @@ func (v *KoanfProvider) PipelineConfig(prefix, id string, override json.RawMessa
 		return errors.WithStack(err)
 	}
```
aeneasr, line 347: The culprit of the high CPU usage is in this method. My recommendation would be to cache the result of the validation (only when successful) and the cache key `sha256(prefix + "|" + id + "|" + marshalled)`. That way we will respect hot reloading (because we use the config values) and still save a lot of CPU cycles!

aeneasr (changes requested):
Thank you very much! I think there's an improvement to this that addresses the problem in a more targeted way and without sideffects :)

David-Wobrock:
Hey @aeneasr 
Thanks for the review! Good catch, we can reduce the caching range, and avoid side-effects 👍 I wasn't aware of the hot-reloading part.

I pushed edits to the PR.

driver/configuration/provider_koanf.go
```diff
@@ -60,13 +59,18 @@ type (
 var _ Provider = new(KoanfProvider)
 
 func NewKoanfProvider(ctx context.Context, flags *pflag.FlagSet, l *logrusx.Logger, opts ...configx.OptionModifier) (kp *KoanfProvider, err error) {
+	cache, _ := ristretto.NewCache(&ristretto.Config{
+		NumCounters: 100,
+		MaxCost:     1 << 22, // 4 MB
+		BufferItems: 64,
```
aeneasr, line 65: This configuration is considered best practice for Ristretto:

```suggestion
		NumCounters:        maxItems * 10,
		MaxCost:            maxItems,
		BufferItems:        64,
		Metrics:            false,
		IgnoreInternalCost: true,
		Cost: func(value interface{}) int64 {
			return 1
		},
```

I would set maxItems to ~5000 which would cover 5000 different configs, more than enough

aeneasr (changes requested):
Awesome, looking much better! just a commend on how to improve the ristretto config

David-Wobrock:
Thanks for the configuration @aeneasr 👍 
I updated the commit with the suggestion :)

aeneasr (approved):

PR #1046, opened 2022-12-29
Fix HEAD not being handled correctly in decision API

jrester:
Currently oathkeeper cannot handle HEAD requests from traefik Forward Auth middleware, because oathkeeper copies the forwarded method into the incoming request. With this fix HEAD requests now also work.

## Related issue(s)

Currently when using the oathkeeper decision API together with traefik forward auth, HEAD requests cannot be handled currectly and will result in a timeout on the traefik side. This is because oathkeeper replaces the method of the incoming request with the method in the `X-Forwarded-Method` header and because HEAD requests *must not* contain a body it will not be written by go (https://github.com/golang/go/blob/9123221ccf3c80c741ead5b6f2e960573b1676b9/src/net/http/server.go#L377). But as traefik sends a `GET` requests for forward auth it expects a body back (https://github.com/traefik/traefik/blob/e54ee89330a800d509da7b11b46a6ecbb331e791/pkg/middlewares/auth/forward.go#L129), therefore traefik times out, as no body is sent by oathkeeper.

### Reproduce
Although not the simplest setup, this is how I noticed this bug:

1. setup oathkeeper with any rule you want (e.g. allow everything)
2. setup the docker registry (https://github.com/distribution/distribution)
3. setup traefik to point to the docker registry with a forward auth middleware to oathkeeper
4. do a docker push with a image of your choice to the docker registry
5. the push should retry multiple times and than fail with a 500 internal Server error
6. the traefik logs should indicate timeout issues

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

## Further Comments

For a more in depth writeup of the problem: https://github.com/thomseddon/traefik-forward-auth/issues/156

### DIFF
--- a/api/decision.go
+++ b/api/decision.go
@@ -5,6 +5,7 @@ package api
 
 import (
 	"net/http"
+	"net/url"
 	"strings"
 
 	"github.com/ory/oathkeeper/pipeline/authn"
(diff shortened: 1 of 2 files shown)


api/decision.go
```diff
@@ -41,13 +42,21 @@ func NewJudgeHandler(r decisionHandlerRegistry) *DecisionHandler {
...
-			x.IfThenElseString(r.TLS != nil, "https", "http"))
-		r.URL.Host = x.OrDefaultString(r.Header.Get(xForwardedHost), r.Host)
-		r.URL.Path = x.OrDefaultString(strings.SplitN(r.Header.Get(xForwardedUri), "?", 2)[0], r.URL.Path[len(DecisionPath):])
-
-		h.decisions(w, r)
+		// Copy the request information, instead of modifing the incoming request directly.
+		// This is nevessary because the middleware would otherwise use e.g. the method from "X-Forwarded-Method" for the response
+		// although the original request had another method, which leads to problem with the HEAD method.
+		// For more information see: https://github.com/thomseddon/traefik-forward-auth/issues/156
+		forwardedReq := &http.Request{
```
aeneasr, line 49: We need to clone the request using `r.Clone()`. If we don't, information is missing that is needed in the decisions function:

- context is missing
- body is missing
- tls information is missing

aeneasr (changes requested):

api/decision.go
jrester: Good point, I updated it

jrester:
@aeneasr are there any more changes required or can this be merged?

8th-block:
This is still an issue it seems. Any updates?!

PR #1049, opened 2023-01-05
refactor: add tracing on hydrator

daviddelucca:
Add tracing into mutator hydrator requests.

## Related issue(s)

https://github.com/ory/oathkeeper/issues/1001

## Checklist

- [X] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md).
- [ ] I have referenced an issue containing the design document if my change
      introduces a new feature.
- [X] I am following the
      [contributing code guidelines](../blob/master/CONTRIBUTING.md#contributing-code).
- [X] I have read the [security policy](../security/policy).
- [X] I confirm that this pull request does not address a security
      vulnerability. If this pull request addresses a security vulnerability, I
      confirm that I got the approval (please contact
      [security@ory.sh](mailto:security@ory.sh)) from the maintainers to push
      the changes.
- [X] I have added tests that prove my fix is effective or that my feature
      works.
- [ ] I have added or changed [the documentation](https://github.com/ory/docs).

## Further Comments

I have removed `http.Client` from Mutator constructor because this is not being utilized at all. Let me know if you want to make any other changes.

I am not adding a design document because this is not a new feature, just added a new tracing point. It's hard to execute troubleshootings if tracing is not displaying all accross a single traceId.

Issue #1001, opened 2022-09-05
Jaeger Tracing: No traceparent header propagation


Poweranimal:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/oathkeeper/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/oathkeeper/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [ ] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe the bug

The Jaeger tracer propagation does not have any effect.
In fact, regardless of any value set for propagation, no tracing information at all are forward to the target by oathkeeper.
In order for tracing to work, oathkeeper must forward tracing information to the target it proxies to (e.g. W3C-Headers, B3-Headers, etc.).

### Reproducing the bug

1. Copy this docker-compose file:
```yaml
services:
  oathkeeper:
    image: docker.io/oryd/oathkeeper:v0.39.3
    entrypoint:
      - ash
      - -ec
    command:
      - |
        echo "$$OATHKEEPER_CONFIG" > /tmp/config.yaml && \
        echo "$$OATHKEEPER_ACCESS_RULES" > /tmp/access-rules.json && \
        oathkeeper serve --config /tmp/config.yaml
    environment:
      OATHKEEPER_CONFIG: |-
        log:
          level: debug
        tracing:
          provider: jaeger
          providers:
            jaeger:
              propagation: b3
              local_agent_address: otc-service:6831
        access_rules: 
          repositories: ["file:///tmp/access-rules.json"]
        authenticators:
          noop: {enabled: true}
        authorizers:
          allow: {enabled: true}
        mutators:
          noop: {enabled: true}
          header:
            enabled: true
            config:
              headers: {}
      OATHKEEPER_ACCESS_RULES: |-
        [
          {
            "id": "hello",
            "upstream": { 
              "url": "http://fake-service:9090"
            },
            "match": {
              "methods": ["GET"],
              "url": "<.*>"
            },
            "authenticators": [
              { "handler": "noop" }
            ],
[...]

PR #1050, opened 2023-01-06
feat: mutator tracing

alnr:
#1001 

This slightly refactors https://github.com/ory/oathkeeper/pull/1049 and keeps more of the timeout behavior unchanged.


pipeline/authn/authenticator_oauth2_introspection_test.go
```diff
@@ -705,7 +705,7 @@ func TestAuthenticatorOAuth2Introspection(t *testing.T) {
 			assertHandlerWasCalled(t)
 
 			t.Run("case=request succeeds and uses the cache", func(t *testing.T) {
-				config := setup(t, `{ "trusted_issuers": ["foo", "bar"], "target_audience": ["audience"] }`)
+				config := setup(t, `{ "required_scope": ["scope-a"], "trusted_issuers": ["foo", "bar"], "target_audience": ["audience"] }`)
```
aeneasr, line 708: This looks like it was an accident or do we need to change this for tracing?

aeneasr (commented):
Just one minor thing for tests otherwise looks good!

pipeline/authn/authenticator_oauth2_introspection_test.go
alnr: Reverted.

alnr:
- updated config.schema.json
- bumped ory/x
- reverted test change (that snuck in during 7c7f21f4128289c25d2421b6fc38f07668483648)

pipeline/mutate/mutator_hydrator.go
```diff
@@ -172,33 +168,28 @@ func (a *MutatorHydrator) Mutate(r *http.Request, session *authn.AuthenticationS
...
-	var client *http.Client
-
+	clientOpts := []httpx.ResilientOptions{httpx.ResilientClientWithTracer(a.d.Tracer())}
 	if cfg.Api.Retry != nil {
-		maxRetryDelay := time.Second
-		giveUpAfter := time.Millisecond * 50
 		if len(cfg.Api.Retry.MaxDelay) > 0 {
+			maxRetryDelay := time.Second
 			if d, err := time.ParseDuration(cfg.Api.Retry.MaxDelay); err != nil {
 				a.d.Logger().WithError(err).Warn("Unable to parse max_delay in the Hydrator Mutator, falling pack to default.")
```
zepatrik, line 176:
```suggestion
				a.d.Logger().WithError(err).Warn("Unable to parse max_delay in the Hydrator Mutator, falling back to default (1s).")
```

pipeline/mutate/mutator_hydrator.go
```diff
@@ -172,33 +168,28 @@ func (a *MutatorHydrator) Mutate(r *http.Request, session *authn.AuthenticationS
...
 	req.Header.Set(contentTypeHeaderKey, contentTypeJSONHeaderValue)
 
-	var client *http.Client
-
+	clientOpts := []httpx.ResilientOptions{httpx.ResilientClientWithTracer(a.d.Tracer())}
 	if cfg.Api.Retry != nil {
-		maxRetryDelay := time.Second
-		giveUpAfter := time.Millisecond * 50
 		if len(cfg.Api.Retry.MaxDelay) > 0 {
+			maxRetryDelay := time.Second
```
zepatrik, line 174: IMO this needs to be moved one scope up, like it was before, so that the default is also set when the config is not set.

pipeline/mutate/mutator_hydrator.go
```diff
@@ -172,33 +168,28 @@ func (a *MutatorHydrator) Mutate(r *http.Request, session *authn.AuthenticationS
...
 			if d, err := time.ParseDuration(cfg.Api.Retry.MaxDelay); err != nil {
 				a.d.Logger().WithError(err).Warn("Unable to parse max_delay in the Hydrator Mutator, falling pack to default.")
 			} else {
 				maxRetryDelay = d
 			}
+			clientOpts = append(clientOpts, httpx.ResilientClientWithMaxRetryWait(maxRetryDelay))
 		}
 		if len(cfg.Api.Retry.GiveUpAfter) > 0 {
+			giveUpAfter := time.Millisecond * 50
```
zepatrik, line 183: Same here.

spec/config.schema.json
```diff
@@ -1915,69 +1915,27 @@
                 "sampling": {
                   "type": "object",
                   "propertyNames": {
-                    "enum": ["type", "value", "server_url"]
+                    "enum": ["server_url", "trace_id_ratio"]
                   },
```
zepatrik, line 1919: This whole key can be removed. Use `additionalProperties: false` instead.
```suggestion
                  "additionalProperties": false,
```

spec/config.schema.json
```diff
@@ -1915,69 +1915,27 @@
...
                   "allOf": [
```
zepatrik, line 1920: The `allOf` can also be removed because there is just one element in the array now.

pipeline/mutate/mutator_hydrator.go
zepatrik: And we need a test for this, so e.g. only setting `cfg.Api.Retry.MaxDelay`, and expecting the `giveUpAfter` default to be set.

zepatrik (changes requested):

pipeline/mutate/mutator_hydrator.go
alnr: After looking at this in more detail, the retry config was completely broken in https://github.com/ory/oathkeeper/commit/2a97e051a98da588aa8125bc0c6681e2d39c48ef.

I'll revisit this...

alnr: Alrighty... the retry configuration is broken all over Oathkeeper. Here's my understanding of what is currently happening:

- The `give_up_after` property specifies the per-retry timeout. That's surprising to me, since it sounds like it means the overall timeout, including retries.
- The `max_delay` specifies the maximum backoff wait time in between retries.
- The maximum number of retries is not configurable, and is always 4 (the default in `hashicorp/retryablehttp`).
- This [config migration](https://github.com/ory/oathkeeper/blob/7150da04e91f515a391e95dd550d48066e25d841/rule/rule_migrator.go#L46-L88) is incorrect. It thinks `give_up_after` is the overall timeout (including retries).

Not sure now is the time to fix this fully, since that could also break users.

With the goal of moving this PR over the line, I propose to keep as much backward-compatibility as possible. This particular mutator previously had no overall timeout and no retries if the retry config was not specified at all. If the retry config was specified (even if `{}`), it peformed up to 4 retries. If the retry config is empty (`{}`), the timeouts also changed.

This patch keeps that behavior, while fixing bugs:
- the fallback timeouts (retry config present but empty) were wrong: 50ms timeout and 1s delay, even though the config schema promises a more reasonable 1s timeout and 100ms delay. Those defaults now at least work as advertised.
- faulty error messages

We should consider revamping the timeout handling all over Oathkeeper in a follow-up ticket.

Thoughts? @zepatrik @aeneasr @daviddelucca ?

spec/config.schema.json
alnr: done, thx

zepatrik (changes requested):
Nice, I agree to merge this as is and fix the other issues you found in a separate PR :+1:

zepatrik (dismissed):

pipeline/mutate/mutator_hydrator.go
daviddelucca: This PR is related to activate tracing on mutator. I think we could merge it and open another ticket/branch to work on revamping on timeouts.

Does make sense?

pipeline/mutate/mutator_hydrator.go
```diff
@@ -172,32 +169,32 @@ func (a *MutatorHydrator) Mutate(r *http.Request, session *authn.AuthenticationS
 	}
 	req.Header.Set(contentTypeHeaderKey, contentTypeJSONHeaderValue)
 
-	var client *http.Client
-
+	client := http.DefaultClient
+	if a.d.Tracer() != nil {
+		client = otelhttp.DefaultClient
+	}
```
aeneasr, line 175: I'm probably late to the party, but the HTTP default client is quite shitty. It doesn't have timeouts etc configured. Why not keep using

```
 		client:       httpx.NewResilientClient().StandardClient(),
```

as the default?

alnr:
This needs more work to synchronize the tracing config schemata with ory/x.

pipeline/mutate/mutator_hydrator.go
alnr: This wasn't the default before this PR. The http.DefaultClient was the fallback in case no retrying was configured.

Let's tackle all the retrying-logic in another PR.

alnr:
I've synced the tracing config schema with the one from ory/x. Initially, I tried to reference it by URL, but that (logically) requires Oathkeeper to load the remote schema on startup, which is often not possible due to network segmentation.

@daviddelucca can you try this branch, please?

aeneasr (approved):

PR #1059, opened 2023-02-03
fix: use watcherx to watch access rule files

zepatrik:


aeneasr (dismissed):
Very nice work, despite the complexity the code is easy to follow :) I couldn't find any logic mistakes or issues, so I think this is 👍 

We should still observe the behavior on a test environment (k3d) to ensure we did not introduce regressions somehow.

zepatrik:
OK, I verified it works with some manual tests in k3d using this manifest:
```yaml
kind: ConfigMap
apiVersion: v1
metadata:
  name: config
data:
  config.yml: |-
    access_rules:
      repositories: 
        - file:///etc/rules/rules.yml
      matching_strategy: regexp
    authenticators:
      anonymous:
        enabled: true
    authorizers:
      allow:
        enabled: true
    mutators:
      noop:
        enabled: true

---

kind: ConfigMap
apiVersion: v1
metadata:
  name: rules
data:
  rules.yml: |-
    - id: test-rule-1-yaml
      upstream:
        preserve_host: true
        strip_path: /api
        url: https://mybackend.com/api
      match:
        url: myproxy.com/api
        methods:
          - GET
          - POST
      authenticators:
        - handler: anonymous
      authorizer:
        handler: allow
      mutators:
        - handler: noop

---

kind: Pod
apiVersion: v1
metadata:
  name: ory-oathkeeper
spec:
  containers:
    - name: oathkeeper
      image: oryd/oathkeeper:dev-alpine
      args:
        - serve
        - -c
        - /etc/configs/config.yml
      ports:
        - containerPort: 4455
        - containerPort: 4456
      env:
        - name: config
          valueFrom:
            configMapKeyRef:
              name: config
              key: config.yml
        - name: rules
          valueFrom:
            configMapKeyRef:
              name: rules
              key: rules.yml
      volumeMounts:
        - name: config
          mountPath: /etc/configs
        - name: rules
          mountPath: /etc/rules
  volumes:
    - name: config
      configMap:
        name: config
    - name: rules
      configMap:
        name: rules
```

First I was confused because the update did not seem to propagate, but turned out it was because of different log levels for config and access rule update logs :sweat_smile:

PR #1061, opened 2023-02-07
fix: rule readiness check should require at least one rule to be loaded

zepatrik:
Currently, all we do is wait for an event to be dispatched. That event is dispatched regardless of whether there actually exists any rule or not. Oathkeeper with zero rules is not very helpful though, so IMO it is a good idea to change this like proposed.


zepatrik:
Hm actually I am not sure if we want to keep https://github.com/ory/oathkeeper/pull/1061/commits/3365b84d97064681825370d7b9e6a9852fe4b039
It changes behavior from 500 on invalid rule match, to 404 :thinking:

api/credential_test.go
```diff
@@ -25,7 +25,11 @@ import (
 
 func TestCredentialsHandler(t *testing.T) {
 	conf := internal.NewConfigurationWithDefaults(configx.SkipValidation())
+	conf.SetForTest(t, configuration.MutatorIDTokenIsEnabled, true)
```
zepatrik, line 28: All these are required because now only valid rules are considered by the rule repository.

cmd/health_alive.go
```diff
@@ -28,7 +28,7 @@ Note:
 	Run: func(cmd *cobra.Command, _ []string) {
 		client := newClient(cmd)
 
-		r, err := client.API.IsInstanceAlive(api.NewIsInstanceAliveParams())
+		r, err := client.Health.IsInstanceAlive(health.NewIsInstanceAliveParams())
```
zepatrik, line 31: I honestly don't know why the SDK changed, but :shrug:

rule/fetcher_default_test.go
```diff
@@ -35,7 +36,7 @@ import (
 	"github.com/ory/oathkeeper/rule"
 )
 
-const testRule = `[{"id":"test-rule-5","upstream":{"preserve_host":true,"strip_path":"/api","url":"mybackend.com/api"},"match":{"url":"myproxy.com/api","methods":["GET","POST"]},"authenticators":[{"handler":"noop"},{"handler":"anonymous"}],"authorizer":{"handler":"allow"},"mutators":[{"handler":"noop"}]}]`
+const testRule = `[{"id":"test-rule-5","upstream":{"preserve_host":true,"strip_path":"/api","url":"https://mybackend.com/api"},"match":{"url":"myproxy.com/api","methods":["GET","POST"]},"authenticators":[{"handler":"noop"},{"handler":"anonymous"}],"authorizer":{"handler":"allow"},"mutators":[{"handler":"noop"}]}]`
```
zepatrik, line 39: A lot of the test rules were actually invalid :sweat_smile:

rule/repository_memory.go
```diff
@@ -133,5 +142,16 @@ func (m *RepositoryMemory) Match(ctx context.Context, method string, u *url.URL,
 		return nil, errors.WithStack(helper.ErrMatchesMoreThanOneRule)
 	}
 
-	return &rules[0], nil
+	return rules[0], nil
+}
+
+func (m *RepositoryMemory) ReadyChecker(r *http.Request) error {
```
zepatrik, line 148: This simple function is so much nicer than that random event pipeline, with actually zero meaning in the end :sweat_smile: but that pre-dated the `ory/x/healthx` package, so the author is excused :wink:

rule/testdata/TestFetchRulesFromObjectStorage-azblob.replay
```diff
@@ -113,7 +113,7 @@
...
...mhhbmRsZXIiOiAiYW5vbnltb3VzIgogICAgICB9CiAgICBdLAogICAgImF1dGhvcml6ZXIiOiB7CiAgICAgICJoYW5kbGVyIjogImRlbnkiCiAgICB9LAogICAgIm11dGF0b3JzIjogWwogICAgICB7CiAgICAgICAgImhhbmRsZXIiOiAiaWRfdG9rZW4iCiAgICAgIH0sCiAgICAgIHsKICAgICAgICAiaGFuZGxlciI6ICJoZWFkZXIiLAogICAgICAgICJjb25maWciOiB7CiAgICAgICAgICAiaGVhZGVycyI6IHsKICAgICAgICAgICAgIlgtVXNlciI6ICJ7eyBwcmludCAuU3ViamVjdCB9fSIKICAgICAgICAgIH0KICAgICAgICB9CiAgICAgIH0KICAgIF0KICB9LAogIHsKICAgICJpZCI6ICJ0ZXN0LXJ1bGUtMyIsCiAgICAidXBzdHJlYW0iOiB7CiAgICAgICJwcmVzZXJ2ZV9ob3N0IjogdHJ1ZSwKICAgICAgInN0cmlwX3BhdGgiOiAiL2FwaSIsCiAgICAgICJ1cmwiOiAiaHR0cHM6Ly9teWJhY2tlbmQuY29tL2FwaSIKICAgIH0sCiAgICAibWF0Y2giOiB7CiAgICAgICJ1cmwiOiAibXlwcm94eS5jb20vYXBpIiwKICAgICAgIm1ldGhvZHMiOiBbCiAgICAgICAgIkdFVCIsCiAgICAgICAgIlBPU1QiCiAgICAgIF0KICAgIH0sCiAgICAiYXV0aGVudGljYXRvcnMiOiBbCiAgICAgIHsKICAgICAgICAiaGFuZGxlciI6ICJub29wIgogICAgICB9LAogICAgICB7CiAgICAgICAgImhhbmRsZXIiOiAiYW5vbnltb3VzIgogICAgICB9CiAgICBdLAogICAgImF1dGhvcml6ZXIiOiB7CiAgICAgICJoYW5kbGVyIjogImFsbG93IgogICAgfSwKICAgICJtdXRhdG9ycyI6IFsKICAgICAgewogICAgICAgICJoYW5kbGVyIjogImlkX3Rva2VuIiwKICAgICAgICAiY29uZmlnIjogewogICAgICAgICAgImp3a3NfdXJsIjogImh0dHA6Ly9zdHViLyIKICAgICAgICB9CiAgICAgIH0KICAgIF0KICB9Cl0="
```
zepatrik, line 116: Just making the rules actually valid...

rule/repository_memory.go
hperl: Agreed!

cmd/health_alive.go
hperl: I think it comes from changes in `github.com/ory/x/healthx`. Fine with me.

hperl (approved):
Great cleanup work. Thanks!

PR #1079, opened 2023-03-14
feat: improved tracing for authorizers

alnr:


zepatrik (approved):
Just FYI, the Keto authorizer is for a very old version of Keto and does not support the latest releases.

aeneasr (approved):

PR #1086, opened 2023-04-03
feat: tracing for gRPC middleware

alnr:
The diff is much smaller when ignoring whitespace.

### DIFF
--- a/middleware/grpc_middleware.go
+++ b/middleware/grpc_middleware.go
@@ -10,6 +10,9 @@ import (
 	"net/url"
 	"strings"
 
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/codes"
+	"go.opentelemetry.io/otel/trace"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/metadata"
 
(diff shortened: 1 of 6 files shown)


middleware/grpc_middleware.go
```diff
@@ -49,85 +52,116 @@ func (m *middleware) httpRequest(ctx context.Context, fullMethod string) (*http.
...
-		if err != nil {
-			log.WithError(err).Warn("failed to handle request")
-			return nil, ErrDenied
-		}
+	httpReq, err := m.httpRequest(traceCtx, info.FullMethod)
+	if err != nil {
+		log.WithError(err).Warn("could not build HTTP request")
+		span.SetAttributes(attribute.String("oathkeeper.verdict", "denied"))
+		span.SetStatus(codes.Error, err.Error())
+		span.End()
```
zepatrik, line 86: Isn't it ended on defer anyway?

zepatrik (changes requested):

middleware/grpc_middleware.go
```diff
@@ -49,85 +52,116 @@ func (m *middleware) httpRequest(ctx context.Context, fullMethod string) (*http.
...
+		return nil, err
```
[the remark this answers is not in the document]
alnr: good catch, this was from a previous iteration

middleware/grpc_middleware.go
alnr: yeah, removed

PR #1091, opened 2023-04-13
fix: noop mutator don't overwrite session headers

DavidSpek:
This PR changes the `noop` mutator so that it doesn't overwrite the current session headers with the request headers.

When using the [remote_json](https://www.ory.sh/docs/oathkeeper/pipeline/authz#remote_json) authorizer, it is possible to configure headers returned by the remote authorizer that will be forwarded to upstream services with the `forward_response_headers_to_upstream` field. This effectively allows the authorizer to also act as a mutator. Since rules must contain a mutator, if no further mutations need to be performed you would specify the `noop` mutator. However, since the `noop` mutator is setting the session headers to be equal to the request headers, this effectively overwrites the headers set by the `remote_json` authorizer.

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

## Further Comments

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1204651423958868

### DIFF
--- a/pipeline/mutate/mutator_noop.go
+++ b/pipeline/mutate/mutator_noop.go
@@ -23,7 +23,22 @@ func (a *MutatorNoop) GetID() string {
 }
 
 func (a *MutatorNoop) Mutate(r *http.Request, session *authn.AuthenticationSession, config json.RawMessage, _ pipeline.Rule) error {
+	currentSessionHeaders := session.Header.Clone()
 	session.Header = r.Header
+	if session.Header == nil {
+		session.Header = make(map[string][]string)
+	}
+
+	for k, v := range currentSessionHeaders {
+		var val string
+		if len(v) == 0 {
+			val = ""
+		} else {
+			val = v[0]
+		}
+		session.SetHeader(k, val)
+	}
+
 	return nil
 }
 
--- a/pipeline/mutate/mutator_noop_test.go
+++ b/pipeline/mutate/mutator_noop_test.go
@@ -33,6 +33,16 @@ func TestMutatorNoop(t *testing.T) {
 		assert.EqualValues(t, r.Header, s.Header)
 	})
 
+	t.Run("method=mutate/case=ensure authentication session headers are kept", func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


pipeline/mutate/mutator_noop.go
```diff
@@ -23,7 +23,22 @@ func (a *MutatorNoop) GetID() string {
...
+	for k, v := range currentSessionHeaders {
+		var val string
+		if len(v) == 0 {
+			val = ""
+		} else {
+			val = v[0]
+		}
+		session.SetHeader(k, val)
+	}
+
```
aeneasr, line 41: Thank you for the fix - the bug you encounter for sure is annoying.

I am trying to figure out what's going on here. So `session.Header` was (I think) originally used by mutators to set a custom header for the upstream. Similarly, the decision endpoint also wants to set these headers (e.g. a JSON Web Token) so that the proxy can forward them to the upstream.

I think that you'll actually have the same problem with other mutators as well, because they are not designed to handle headers from the authorization piece.

It probably makes sense to manage merging the two headers from the incoming request and the session outside of the handlers?

And it definitely needs a failing test case first to ensure this fix works and doesn't regress.

DavidSpek: Sorry for not getting back to your review quickly. I believe I looked at the code for the other mutators and they didn't have that problem. I'll have another look and see if I can find an elegant way to handle it outside of the handlers if that makes sense. Initially I made this PR as small as possible so it would be easy to review. I've deployed this fix already so I know it works, but a test would be good to add to ensure it doesn't regress in the future.

DavidSpek: @zepatrik Maybe you can way in here as well.

DavidSpek:
Ping @aeneasr @zepatrik

aeneasr (approved):
Sorry for the long review, this LGTM!

DavidSpek:
No problem at all :)

Issue #1105, opened 2023-05-24
Authenticator: Bearer_token w. "query_parameter" selector consumes request body


marbergq:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/oathkeeper/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/oathkeeper/blob/master/CONTRIBUTING.md).
- [X] This issue affects my [Ory Network](https://www.ory.sh/) project.
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [X] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe the bug

GIven this rule
```
      authenticators:
      - handler: bearer_token
        config:
          check_session_url: http://check-session-host/...
          token_from:
            query_parameter: token
```
The upstream service will never see the body when the request is of `Content-type: application/x-www-form-urlencoded`.

This is because `return r.FormValue(*tokenLocation.QueryParameter)` (in `helper/bearer.go`) will consume the body 

### Reproducing the bug

config.yaml: 
```serve:
  api:
    port: 6061
  proxy:
    port: 6060

access_rules:
  repositories:
    - file://./rules.1.json

authenticators:
  noop:
    enabled: true
  bearer_token:
    enabled: true
    config:
      check_session_url: http://localhost:6662/session

authorizers:
  allow:
    enabled: true
  deny:
    enabled: true

mutators:
  noop:
    enabled: true
```

rules.1.yaml

```
[
  {
    "id": "test",
    "upstream": {
      "url": "http://127.0.0.1:6662"
    },
    "match": {
      "url": "http://127.0.0.1:6060/test",
      "methods": ["POST"]
    },
    "authenticators": [
      {
        "handler": "bearer_token",
        "config": {
          "check_session_url": "http://127.0.0.1:6662/session",
          "preserve_path": true,
          "preserve_query": false,
          "force_method": "GET",
          "token_from": {
            "query_parameter": "token"
          }
        }
      }
    ],
    "authorizer": {
      "handler": "allow"
    },
    "mutators": [
      {
        "handler": "noop"
      }
    ]
  }
]
```

run.sh

```
#!/bin/bash

[...]

PR #1106, opened 2023-05-24
fix: [Bearer_token] Use Query.Get when fetching QueryParameter

marbergq:
Fix for [bug](https://github.com/ory/oathkeeper/issues/1105) where the bearer_token authn consumes body.
Instead of using `r.FormValue` which parses body (and thus consumes it, so it's not available to upstream services), we can just use `r.URL.Query().Get(*tokenLocation.QueryParameter)`


aeneasr (approved):

PR #1124, opened 2023-07-26
chore: bump ory/herodot

alnr:
Also minor dependency cleanup. This should help with client-canceled requests showing up as 500 errors in logs. They should now be reported as 499s.


.prettierignore
```diff
@@ -0,0 +1 @@
+.schema/version.schema.json
```
zepatrik, line 1: Why this?

alnr: ory/ci and the `format` make target in this repo disagree about the number of spaces to indent with 🤮

aeneasr (approved):
LGTM

PR #1134, opened 2023-09-07
fix: avoid unescaping slashes when proxying URLs

refi64:
Doing all the path operations on URL.Path means that they're being performed on an unescaped URL. Unfortunately, this URL is *not* guaranteed to be the directly unescaped version of URL.EscapedPath(); in particular, slashes will be unescaped in URL.Path, e.g. given:

/abc%2Fdef

URL.Path will be:

/abc/def

In these cases, URL.RawPath is also set with the escaped version of the path. However, once URL.Path is modified, URL.RawPath is ignored, and URL.EscapedPath() returns the path-escaped version of URL.Path...which, in this case, is now /abc/def, not the correct /abc%2Fdef.

As a result, when performing any path modifications during proxying (i.e. proxying to an upstream URL with a path component, or applying StripPath), this results in any *escaped* slashes in the proxied URL being unescaped.

In order to work around this, rewriting operations need to take place on the *escaped* path, not the unescaped one, and then an intermediate URL can be used to determine the Path and RawPath values to set on the forward URL.

This incurs a small breaking change in that StripPath is now applied to the *escaped* URL, not the unescaped URL. These semantics arguably make more sense, since StripPath otherwise also cannot distinguish between unencoded slashes within a path segment vs those that are separating path segments, but it's still a breaking change nonetheless.

```
BREAKING CHANGES: This patch changes the behavior of strip_path to be applied to the *escaped* URL as returned by Go's `URL.EscapePath()`, which means that slashes inside path segments, carets, unicode characters, and others that are need to be percent-encoded in strip_path.
```

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
  - (the listed email was contacted in advance and confirmed that this change can be opened as a public PR)
      works.

## Further Comments

I debating trying to preserve the current `strip_path` semantics, but in addition to being unable to distinguish `/` and `%2F`, it's also just *incredibly* ugly: I need to put the value of the attribute in a stub URL and parse that to get the escaped path, which also introduces a large amount of weird edge cases that come from quirks in URL parsers.

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -184,20 +184,46 @@ func ConfigureBackendURL(r *http.Request, rl *rule.Rule) error {
 	}
 
 	proxyHost := r.Host
-	proxyPath := r.URL.Path
+	proxyPath := r.URL.EscapedPath()
 
 	backendHost := p.Host
-	backendPath := p.Path
+	backendPath := p.EscapedPath()
 	backendScheme := p.Scheme
+
+	// We need to build a new path from the incoming *escaped* path components,
+	// because just relying on the already-escaped URL.Path results in escaped
+	// slashes (%2F) being irreversibly converted to actual slashes.
+	//
+	// However, once the new escaped path is built, we can't immediately assign it
+	// to the resulting URL:
+	// - Assigning to URL.Path would mean assigning an escaped path to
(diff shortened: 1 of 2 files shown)


em-:
Any chance to get this reviewed? Thank you!

Issue #1136, opened 2023-09-09
Authorizer "remote" throws exception "invalid Read on closed Body" if request body is present in request


denysandriyanov:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/oathkeeper/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/oathkeeper/blob/master/CONTRIBUTING.md).
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [X] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe the bug

I am using the authorizer "remote". 
When sending a request with some JSON payload and the remote authorizer returns status code 200 it is expected that Oathkeeper will allow the request and forward it to upstream service, while the actual result is I am getting 502 status code. In oathkeeper logs, i see the exception "invalid Read on closed Body".

The authorizer implementation that is used during this flow accepts POST request with a body and returns status code 200 or 403 based on body content. 
Authorizer do not return any body in response, just a status code.

### Reproducing the bug

1. Configure oathkeeper to use authorizer "remote"
2. Configure a dummy API service (mocking behavior of authorizer) that accepts POST request with some json payload as body and returns status code 200
3. Create oathkeeper rule to protect some api endpoint and use the configured authorizer 
4. Make a call to the endpoint configured in step 3

Expected result:
Considering that our dummy authorizer returned status code 200 oathkeeper is expected to allow the request to upstream service

Actual result:
Exception is happened in oathkeeper indicating that  "invalid Read on closed Body". 

### Relevant log output

[...]

PR #1138, opened 2023-10-12
fix: Authorizer "remote" throws exception

timthornton-avid:
…ed Body" if request body is present in request

https://github.com/ory/oathkeeper/issues/1136

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.


pipeline/authz/remote.go
```diff
@@ -67,7 +67,24 @@ func (a *AuthorizerRemote) GetID() string {
...
+	// The calling function expects the original request message to be modified as the message is processed by the pipeline.
+	// When message processing is complete the calling function does not have access to the modified http message (r)
+	//
+	// There are several possible solutions:
+	// 1. Create a function in http.request to .SetContext(ctx) which replaces context but does not create new http.Request instance
+	// 2. Modify the Authorize function to return a new http.Request pointer used by this function during message processing
+	// 3. Assign the new request to the original request pointer
+	//
+	// The change below implements option 3
+	*r = *(r.WithContext(ctx))
```
taisph, line 86: I'm probably wrong but this looks odd to me. `WithContext` creates a shallow clone of the request which might be the actual issue, given how `r` is used below. Have you tried `r = r.Clone(ctx)` instead?

alnr:
Thanks for the PR! Would you mind adding a failing test case as a first step? That way we're sure to catch any regressions in the future. Also, it makes it easier to understand the problem and convince ourselves the fix is correct.

pipeline/authz/remote.go
timthornton-avid: Hi taishp.. Yes I tried Clone() also. Using either function r.WithContext(ctx) or r.Clone(ctx) create a new request message which gets assigned to r (HTTP Request *). The calling function to Authorizer expects r* it passed to persist throughout the pipeline as it will use the context and payload for further processing. Once we assign r (via WithContext or Clone) in the Authorize function, it's pointing to a new HTTP Request * which goes out of scope when the Authorize function returns. 

It appears this issue was introduced in #1079 when adding tracing improvements. 
While testing, if we remove the lines added from #1079:
	ctx, span := a.tracer.Start(r.Context(), "authz.remote")
	defer otelx.End(span, &err)
	r = r.WithContext(ctx)
The issue does not occur as the scope of r does not change.

As another test, I added a function SetContext(ctx) to http.request to assign the context to the request messaeg but not create a new http.http request. This also worked as again the scope of r persisted.

timthornton-avid: To address the syntax looking odd, here is another datapoint that follows the same pattern as this code change. Line 133 calls HandleRequest which calls the remote::Authorize function where the fix was made. In the below code snippet we confirm that the r.Out scope must be retained throughout the message processing.

proxy.go 
132	*r.Out = *r.Out.WithContext(context.WithValue(r.Out.Context(), ContextKeyMatchedRule, rl))
133	s, err := d.r.ProxyRequestHandler().HandleRequest(r.Out, rl)
134	if err != nil {
135		*r.Out = *r.Out.WithContext(context.WithValue(r.Out.Context(), director, err))
136		return
137	}
138	*r.Out = *r.Out.WithContext(context.WithValue(r.Out.Context(), ContextKeySession, s))

timthornton-avid: Just enhanced and added to pull the unit testing in remote_test.go. To test this failure the message body sent to the authorize function is compared to the message body after the authorize pipeline completes. There are two tests that already pass a message body into the Authorize function and these were both verified to fail without 1136 fix and pass after 1136 fix.

ouranders:
@timthornton-avid Really nice of you to fix this. Could you get the branch up-to-date with the base branch and try to resolve the failing "Docker Image Scanners"? We would really love to have this issue fixed.
Thanks!

ralfkrakowski:
Any progress on this pull? We would really love to have this issue fixed.
Thank you!

timthornton-avid:
Hi, I'll try to get this branch up to date and retest this week or next. As per the docker image failure, there was nothing I added that would impact docker image. There was a 1 line code change + comments and a unit test added. I don't recall the docker image scan failure but they would have existed prior to my changes. Lets see what it looks like after I rebase.

ralfkrakowski:
Hi, is there any progress with the review of this issue @taisph and @aeneasr ? 
This fix would be really valuable for us. 
Thanks for releasing such a great product!

timthornton-avid:
Hey Guys, 
Any update on this ?  Updated branch a month ago.

keyur-scogo:
Hi any plans on releasing this fix?

ouranders:
As others has previously mentioned, we are also very eager to have this issue fixed and merged :) @taisph @aeneasr any news on when this can be looked on? Kudos for a great product!

alnr:
Fixed in https://github.com/ory/oathkeeper/issues/1136. Thanks for the PR and your patience.

Issue #1089, opened 2023-04-11
Allow for easily matching rules using path prefixes


davidspek:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/oathkeeper/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/oathkeeper/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Network](https://www.ory.sh/) project.
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Context and scope

Currently rules are matched to an incoming request based on regex or glob pattern matching. Since only a single rule is allowed to match a request, the regex or glob patterns must be very precise and become increasingly complex as different paths need to be matched. While the used regex library supports negative lookahead, creating rules where requests are matched based on path prefixes is still difficult since you'd need to provide negative matches for all other rule regexes with the same base path. As glob doesn't support negative lookahead doing path prefix based routing is not possible. Even when using regex patterns with negative lookahead, the issue that arises is that the end user ends up needing to manage a system with the state of all rules to then be able to create rules with the appropriate regexes including negative lookahead patterns for all other rules. Since the most common type of routing people likely would want to implement is prefixed based routing, the fact that doing so with oathkeeper is difficult and error prone is in my views its biggest drawback.

Also see https://github.com/ory/oathkeeper/issues/441.

### Goals and non-goals

Goals:

- Allow for easily matching routing rules using path prefixes.

Non-goals:

- Allow for using regex in the scheme section of the URL, since that would make it much more difficult to enter rules into the Trie.
-  Allow for regex or glob patterns in places other in the host section of the URL, since making a Trie compatible with pattern matching is likely very difficult if not impossible.
- Allow for regex or glob patterns in places other than the last section of the path, since making a Trie compatible with pattern matching is likely very difficult if not impossible.

### The design

[...]

PR #1147, opened 2024-01-31
fix: update regex engine to support possessive match and lookbehind syntaxes

cmmoran:
Update the regexp to conditionally (naively?) change the current start,end delimiters for the regexp from the hard-coded '<','>' to '{','}' when the compiler encounters a pattern that has syntax that includes '(?<' or '(?>' organically.

BREAKING CHANGES: None

## Related issue(s)

Relates:
#441

Fixes:
#1089

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/rule/engine_regexp.go
+++ b/rule/engine_regexp.go
@@ -4,8 +4,10 @@
 package rule
 
 import (
-	"errors"
 	"hash/crc64"
+	"strings"
+
+	"github.com/pkg/errors"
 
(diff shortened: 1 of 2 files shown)


aeneasr:
Thank you for the changes, however the reason lookbehind is not working is because we're using the golang std library, which does not support these features.

aeneasr:
I'm actually wrong, looksl ike this is working, which is very nice!

PR #1160, opened 2024-04-24
feat(cmd/serve): Add default shutdown timeout

ottingbob:
Adds a default shutdown timeout to help us work with our remix / converge applications.

Also adds in CI to push the image to our respective container registries


ottingbob:
This was not staged correctly to be merged uptream as it had some custom CI

PR #1162, opened 2024-04-29
feat(cmd/serve): add in default-shutdown-timeout flag to increase shutdown timeout on http server shutdown

ottingbob:
Under the hood `ory/oathkeeper` uses https://github.com/ory/graceful to shutdown the HTTP proxy. The `graceful` package provides a constant called `DefaultShutdownTimeout` which sets the timeout to be `5` seconds.

We have an application that can take up to `60` seconds to finish processing requests and we need a solution to be able to change this shutdown time in order for those requests to complete successfully.

As a result, we need a way to configure `oathkeeper` to bump up that timeout and the implementation has a flag that will be able to override the global variable set in the `graceful` package. Although this does not seem like the most **ideal** solution it provides a way for us to change that value.

Curious to hear suggestions on what may be an alternative approach to get a similar outcome & would be happy to try and make the changes needed for any feedback to help enable that.

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.

### DIFF
--- a/cmd/serve.go
+++ b/cmd/serve.go
@@ -32,4 +32,5 @@ func init() {
 
 	serveCmd.PersistentFlags().Bool("disable-telemetry", false, "Disable anonymized telemetry reports - for more information please visit https://www.ory.sh/docs/ecosystem/sqa")
 	serveCmd.PersistentFlags().Bool("sqa-opt-out", false, "Disable anonymized telemetry reports - for more information please visit https://www.ory.sh/docs/ecosystem/sqa")
+	serveCmd.PersistentFlags().Int("default-shutdown-timeout", 5, "Set the default shutdown timeout in seconds for server shutdown when trapping SIGTERM and SIGINT")
 }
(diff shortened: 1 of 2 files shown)


cmd/serve.go
```diff
@@ -32,4 +32,5 @@ func init() {
 
 	serveCmd.PersistentFlags().Bool("disable-telemetry", false, "Disable anonymized telemetry reports - for more information please visit https://www.ory.sh/docs/ecosystem/sqa")
 	serveCmd.PersistentFlags().Bool("sqa-opt-out", false, "Disable anonymized telemetry reports - for more information please visit https://www.ory.sh/docs/ecosystem/sqa")
+	serveCmd.PersistentFlags().Int("default-shutdown-timeout", 5, "Set the default shutdown timeout in seconds for server shutdown when trapping SIGTERM and SIGINT")
```
aeneasr, line 35: In my view this should go into the config file instead

aeneasr (commented):

Issue #1169, opened 2024-06-15
Match response body in remote_json authorizer


jaspeen:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/oathkeeper/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/oathkeeper/blob/master/CONTRIBUTING.md).
- [ ] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [X] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Context and scope

There are external autorization services like [open policy agent](https://www.openpolicyagent.org/) and [OpenFGA](https://openfga.dev/) which don't response with HTTP 403 on authorization failure. Instead they return json value indicating the result of authorization - `{"result": true}` and `{"allowed": true}` for OPA and OpenFGA correspondingly.

In this case we need to check the response body to determine if the request is authorized or not.

There is related issue #1125 

### Goals and non-goals

- Deny or allow request based on the particular response json body value in case response status is 200
- Backward compatible with existing configuration

### The design

New configuration key will be added under `remote_json` authorizer - `match_json_field` as object with following properties:
- `field` : json field selector (gjson) to match
- `str_val` : string value to match
- `bool_val` : boolean value to match

If that key is specified in authorizaer configuration and reponse HTTP status is 200 it will try to parse the response body as json, extract the field value using gjson selector and compare it with the provided value.
 In this case authorization passed only if the field exists and equals with value of proper type.

From my point of view gjson selector is better here than just simple parse json and get the field because beyond those mentioned external authorization service there may be other which put the result in nested structure like `{"result":{"status": "ok"}}` and selector gives there more flexibility. And gjson is already used in the codebase.

Issue #1164, opened 2024-05-14
Oathkeeper docker-compose.yaml outdated env vars


cerealkill:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/oathkeeper/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/oathkeeper/blob/master/CONTRIBUTING.md).
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe the bug

Oathkeeper default [docker-compose.yml](https://github.com/ory/oathkeeper/blob/master/docker-compose.yml#L13C1-L17C49) have outdated metric variables  

### Reproducing the bug

```
git clone git@github.com:ory/oathkeeper.git
cd oathkeeper
docker compose up
```

### Relevant log output

```shell
oathkeeper                  | {"audience":"application","error":{"message":"missing port in address","stack_trace":"stack trace could not be recovered from error type *net.AddrError"},"level":"fatal","msg":"Unable to initialize Tracer for Oathkeeper.","service_name":"ORY Oathkeeper","service_version":"v0.40.7","time":"2024-05-14T09:19:08.243127416Z"}
```

### Relevant configuration

```yml
services:
  oathkeeper:
    environment:
      - TRACING_PROVIDER=jaeger
      - TRACING_PROVIDER_JAEGER_SAMPLING_SERVER_URL=http://jaeger:5778/sampling
      - TRACING_PROVIDER_JAEGER_LOCAL_AGENT_ADDRESS=jaeger:6831
      - TRACING_PROVIDER_JAEGER_SAMPLING_TYPE=const
      - TRACING_PROVIDER_JAEGER_SAMPLING_VALUE=1
```

### Version

0.40.7

### On which operating system are you observing this issue?

Linux

### In which environment are you deploying?

Docker Compose

### Additional Context

Env vars changed, to fix just change to:
```
services:
  oathkeeper:
    environment:
      - TRACING_PROVIDER=jaeger
      - TRACING_PROVIDERS_JAEGER_SAMPLING_SERVER_URL: http://jaeger:5778/sampling
[...]

PR #1170, opened 2024-06-15
feat: match_json_field in remote_json authorizer (#1164)

jaspeen:
Implements response content match in remote_json authorizer as described in rfc #1169 

## Related issue(s)
#1169

## Checklist

      introduces a new feature.
- [ ] I am following the
      the changes.
      works.

### DIFF
--- a/pipeline/authz/remote_json.go
+++ b/pipeline/authz/remote_json.go
@@ -8,11 +8,13 @@ import (
 	"crypto/sha256"
 	"encoding/json"
 	"fmt"
+	"io"
 	"net/http"
 	"text/template"
 	"time"
 
 	"github.com/pkg/errors"
+	"github.com/tidwall/gjson"
 
 	"github.com/ory/x/httpx"
 	"github.com/ory/x/otelx"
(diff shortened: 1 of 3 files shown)


aeneasr (changes requested):
Thank you for your contributino and work on this feature! Unfortunately, I feel that the work can not be merged to upstream. I think there are two topics at play here:

1. Oathkeeper assumes that the upstream gives the correct response it expects (versus it must interpret the response)
2. We have an explicit contract that says "yes ok" or "no not ok". As soon as we go into dynamic parsing this becomes muddied
3. In general I think that the evaluation logic is thin. It probably would make more sense to use Google CEL or something like that. But that again is absolute overkill in my view.

Honestly, I don't see a clear way forward with this PR because I don't really think that we should allow dynamic matching here as it is - in my view - easy to misconfigure.

jaspeen:
@aeneasr Thank's for your review.
I understand your concerns and agree that degree of freedom in such things should be minimal but please look from the user(sre/devops) perspective. 
Some external ready-to-go authorizers don't return HTTP 403 and they require some sort of glue to be integrated into an oathkeeper pipeline. 
And such kind of dynamic integration glue should be executed somewhere - in oathkleeper itself or custom integration service which then return 403. From my point of view another integration service may increase chance of error or misconfiguration.

I also understand why RP's like envoy or nginx don't bother with authorization response checks - they focused on other things. But for oathkeeper as auth decision engine i see it quite natural.

We can minimize flexibility of this and just check a top level boolean JSON field with specified name in response - e.g. `"allow_json_field": "allowed"`. It will cover both mentioned external authorizer cases. and will explicitly declare that authorization passes if and only if response is successfull and field `allowed` has value true.

What do you think?

aeneasr:
To be honest, I'm undecided on this topic.

* As discussed prior there currently no real request transformation in the authn / authz pipeline (only mutator) right now. We forward the request as is to the upstream.
* Adding functionality to "generally" support more upstream servers does makes sense. The upstream server not responding with a 4xx error is a very interesting choice in my view though.
* How far does this go? Do we support XML and others as well? How complex should rules be?

To me, right now, `allow_json_field` seems like a solution specific to one or two providers that probably isn't applicable for a big enough chunk (assuming that we might also get other fields returned like a scope or whatever).

jaspeen:
> currently no real request transformation in the authn / authz

Not exactly. We have a request payload transformation using jsonnet for `remote_json`. But my point is actually that this is not a transformation but matching. Here we already doing kind of a matching of remote authorizer HTTP status with 200 and this is going to be an enhanced version of response matching.

Regarding more 'general' solution for matching i think JSON will cover most of the cases with modern software which accessible via HTTP REST (gRPC and so on are currently out of scope anyway for all rules). And we can also use `jsonnet` for sophisticated multifield matching. Since i see already a lot of expression in `gotemplate` and `jsonnet` so i think adding another one is not the best idea.

Anyway when you get any certain idea how to do this properly i will be glad to adopt this solution.

PR #1177, opened 2024-08-09
feat: make id_token mutator cache configurable

David-Wobrock:
Make the `id_token` mutator cache configurable:
- can be enabled/disabled
- can set the `max_cost`

Changes to default configuration:
Previous:
- NumCounters: 10000
- MaxCost: 1 << 25

New:
- NumCounters: maxCost * 10
- MaxCost: 1 << 25
- Cost function: JWT length

## Related issue(s)

Follow up of https://github.com/ory/oathkeeper/pull/1171 and https://github.com/ory/oathkeeper/pull/1209 (and https://github.com/ory/oathkeeper/pull/1210 too a bit).

Related docs PR: https://github.com/ory/docs/pull/1820

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

## Further Comments

Could probably be subject to a minor version bump, since there's a behaviour change.

## Summary by CodeRabbit

* **New Features**
  * Added per-rule caching controls for ID token generation.
  * Cache settings include an enable/disable option, enabled by default, and a configurable maximum cost of 33,554,432 to manage memory usage.
  * Added configuration support across the ID token mutator schemas.
* **Tests**
  * Added coverage for disabled caching and entries exceeding the configured maximum cost.

### DIFF
--- a/.schema/config.schema.json
+++ b/.schema/config.schema.json
@@ -1140,6 +1140,25 @@
           "pattern": "^[0-9]+(ns|us|ms|s|m|h)$",
           "default": "15m",
           "examples": ["1h", "1m", "30s"]
+        },
+        "cache": {
+          "additionalProperties": false,
+          "type": "object",
+          "properties": {
+            "enabled": {
+              "title": "Enabled",
+              "type": "boolean",
+              "default": true,
+              "examples": [false, true],
+              "description": "En-/disables this component."
+            },
+            "max_cost": {
+              "type": "integer",
+              "default": 33554432,
+              "title": "Maximum Cached Cost",
+              "description": "The cost of one cached JSON Web Token is the length of its string form."
+            }
+          }
         }
       },
       "additionalProperties": false
--- a/pipeline/mutate/mutator_id_token.go
+++ b/pipeline/mutate/mutator_id_token.go
@@ -38,15 +38,20 @@ type MutatorIDToken struct {
 	templates     *template.Template
 	templatesLock sync.Mutex
 
-	tokenCache        *ristretto.Cache[string, *idTokenCacheContainer]
-	tokenCacheEnabled bool
+	tokenCache *ristretto.Cache[string, *idTokenCacheContainer]
 }
 
 type CredentialsIDTokenConfig struct {
-	Claims    string `json:"claims"`
-	IssuerURL string `json:"issuer_url"`
-	JWKSURL   string `json:"jwks_url"`
-	TTL       string `json:"ttl"`
+	Claims    string             `json:"claims"`
+	IssuerURL string             `json:"issuer_url"`
+	JWKSURL   string             `json:"jwks_url"`
+	TTL       string             `json:"ttl"`
+	Cache     IdTokenCacheConfig `json:"cache"`
+}
+
+type IdTokenCacheConfig struct {
+	Enabled bool `json:"enabled"`
+	MaxCost int  `json:"max_cost"`
 }
 
 func (c *CredentialsIDTokenConfig) ClaimsTemplateID() string {
@@ -54,12 +59,7 @@ func (c *CredentialsIDTokenConfig) ClaimsTemplateID() string {
 }
 
 func NewMutatorIDToken(c configuration.Provider, r MutatorIDTokenRegistry) *MutatorIDToken {
-	cache, _ := ristretto.NewCache(&ristretto.Config[string, *idTokenCacheContainer]{
-		NumCounters: 10000,
-		MaxCost:     1 << 25,
-		BufferItems: 64,
-	})
-	return &MutatorIDToken{r: r, c: c, templates: x.NewTemplate("id_token"), tokenCache: cache, tokenCacheEnabled: true}
+	return &MutatorIDToken{r: r, c: c, templates: x.NewTemplate("id_token")}
 }
 
 func (a *MutatorIDToken) GetID() string {
@@ -68,10 +68,6 @@ func (a *MutatorIDToken) GetID() string {
 
 func (a *MutatorIDToken) WithCache(t *template.Template) {
 	a.templates = t
-}
-
-func (a *MutatorIDToken) SetCaching(token bool) {
-	a.tokenCacheEnabled = token
 }
 
 type idTokenCacheContainer struct {
@@ -86,7 +82,7 @@ func (a *MutatorIDToken) cacheKey(config *CredentialsIDTokenConfig, ttl time.Dur
 }
 
 func (a *MutatorIDToken) tokenFromCache(config *CredentialsIDTokenConfig, session *authn.AuthenticationSession, claims []byte, ttl time.Duration) (string, bool) {
-	if !a.tokenCacheEnabled {
+	if !config.Cache.Enabled {
 		return "", false
 	}
 
@@ -106,7 +102,7 @@ func (a *MutatorIDToken) tokenFromCache(config *CredentialsIDTokenConfig, sessio
 }
 
 func (a *MutatorIDToken) tokenToCache(config *CredentialsIDTokenConfig, session *authn.AuthenticationSession, claims []byte, ttl time.Duration, expiresAt time.Time, token string) {
-	if !a.tokenCacheEnabled {
+	if !config.Cache.Enabled {
 		return
 	}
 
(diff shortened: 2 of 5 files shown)


pipeline/mutate/mutator_id_token.go
```diff
@@ -204,5 +204,29 @@ func (a *MutatorIDToken) Config(config json.RawMessage) (*CredentialsIDTokenConf
...
+		cache, err := ristretto.NewCache(&ristretto.Config{
+			NumCounters: cost * 10,
+			// Allocate a max
+			MaxCost: cost,
+			// This is a best-practice value.
+			BufferItems: 64,
+			Cost: func(value interface{}) int64 {
+				return 1
+			},
+			IgnoreInternalCost: true,
```
David-Wobrock, line 222: To be discussed if this is really the best strategy for `id_token`.

I'm less familiar with other use cases of generated `id_token` and also how `ristretto` computes the "cost".

For now, this is copied from the AuthN OAuth2 introspection handler.
But perhaps saying that each id_token = 1 cost would cache too many id_tokens.

I'd like @aeneasr's opinion here :)

David-Wobrock: In the current state it does not work, but we could at least do
```
                        NumCounters: 10000,
			BufferItems: 64,
                        MaxCost: cost,
```
to keep the same behaviour as previously.

aeneasr: cost = 1 << 25

as the default is definitely too many now that the cost function is 1

aeneasr: Since JWTs can be quite long it may make more sense to calculate the cost based on the string length to have an approximation of storage use

David-Wobrock: That could make sense indeed 👍 
Then the default limit of 1 << 25 (~33.5 million), if we estimate an average length of JWT of a few thousands char, we should be able to store some tens of thousands of keys - which sounds reasonable.

pipeline/mutate/mutator_id_token.go
```diff
@@ -108,16 +104,21 @@ func (a *MutatorIDToken) tokenFromCache(config *CredentialsIDTokenConfig, sessio
...
+	a.tokenCache.SetWithTTL(
+		key,
+		&idTokenCacheContainer{
+			TTL:       ttl,
+			ExpiresAt: expiresAt,
+			Token:     token,
+		},
+		0,
+		ttl,
+	)
```
David-Wobrock, line 121: Another probably reason for the memory leak:

Keys are only deleted from cache when requested and expired.

I don't know if ristretto flushes other keys when the max cost is reached, but this could be another lead.

David-Wobrock: We deployed this in production, and it fixed the memory leak indeed :)

spec/config.schema.json
```diff
@@ -1140,6 +1140,25 @@
           "pattern": "^[0-9]+(ns|us|ms|s|m|h)$",
           "default": "15m",
           "examples": ["1h", "1m", "30s"]
+        },
+        "cache": {
+          "additionalProperties": false,
+          "type": "object",
+          "properties": {
```
alnr, line 1147: We have a bunch of other caches in the config schema already. Can you make your change so that it is more similar (identical) to those other cache configurations? The `max_cost` parameter in particular is really opaque and its impossible to come up with a reasonable value without knowing the implementation.

pipeline/mutate/mutator_id_token.go
```diff
@@ -108,16 +104,21 @@ func (a *MutatorIDToken) tokenFromCache(config *CredentialsIDTokenConfig, sessio
...
-	}, 0)
+	a.tokenCache.SetWithTTL(
+		key,
+		&idTokenCacheContainer{
+			TTL:       ttl,
+			ExpiresAt: expiresAt,
+			Token:     token,
+		},
+		0,
+		ttl,
```
alnr, line 120: In other caches, I believe we set the TTL to `min(TTL, time.Until(expiresAt)`. That would make sense here too IMO.

pipeline/mutate/mutator_id_token.go
```diff
@@ -70,10 +70,6 @@ func (a *MutatorIDToken) WithCache(t *template.Template) {
 	a.templates = t
 }
 
-func (a *MutatorIDToken) SetCaching(token bool) {
-	a.tokenCacheEnabled = token
-}
-
 type idTokenCacheContainer struct {
 	ExpiresAt time.Time
```
alnr, line 74: I haven't looked at this code in detail, but `ExpiresAt`+`TTL` seem to be semantically the same thing? Do we even need to store these in the cache?

alnr (changes requested):

pipeline/mutate/mutator_id_token.go
David-Wobrock: They should be same normally. Since calling this method does:
```
	now := time.Now().UTC()
	exp := now.Add(ttl)
	[...]
	a.tokenToCache(c, session, templateClaims, ttl, exp, signed)
```

Or what am I missing? :)

spec/config.schema.json
David-Wobrock: I agree with the remark on `max_cost` 👍 
A value with similar semantics is used today for the OAuth2 introspection authenticator, so I mainly mimicked this one.
Else, there is also `max_tokens`, used by the OAuth2 client credentials authenticator.

I think the reason for the `cost` instead comes from the fact that the cached objects have variable lengths, so storing a certain number of objects will result in a different cache memory usage depending on your config.

However, one can also make the decision to let the user make this decision anyway :)
Let me know what makes most sense to you, and Ory's strategy around these questions (should the product have the same defaults for everyone, or is the user trusted to configure this accordingly).

----

For the `enabled/disabled` value, I didn't re-use the existing
```
    "handlerSwitch": {
      "title": "Enabled",
      "type": "boolean",
      "default": false,
      "examples": [true],
      "description": "En-/disables this component."
    },
```
to avoid introducing a breaking change.

Currently, the `id_token` cache is enabled by default, and didn't wanna change the default value to false - so I couldn't re-use this configuration.

---

And finally, this cache config has no TTL, because the id_token mutator already has a TTL config value for the JWT expiration date.
It make sense to me to re-use the same value => cache for 15 min if the JWT is valid 15min.

pipeline/mutate/mutator_id_token.go
David-Wobrock: Good catch, we don't!

I think that's what was buggy in the first place: we don't set an actual TTL in the cache, but instead cache the TTL along with the token.

I pushed edit 👍

pipeline/mutate/mutator_id_token.go
aeneasr: LGTM

spec/config.schema.json
aeneasr: Ideally, we would probably have one namespaced cache for all of oathkeeper and then use that everywhere, but this is good for now!

aeneasr (commented):
This only changes cache sizes and not the actual caching function itself, right? If so I think we’re very close!

David-Wobrock:
> This only changes cache sizes and not the actual caching function itself, right? If so I think we’re very close!

I'm glad to read this 😁 

At the time of writing this patch does:
- behaviour changes:
  - set a TTL on keys in id_token mutator cache
  - set a cost function in the id_token mutator cache
- new configuration:
  - allow disabling id_token mutator cache
  - allow changing the cost of id_okten mutator cache

David-Wobrock:
Hey @aeneasr I hope you're well :)

Did you get a chance to have a look again? 😇 
We are still running our forked and patched Oathkeeper, but we would obviously prefer running the Ory upstream version.

Perhaps this can make it into the next version?

David-Wobrock:
Hello @aeneasr @alnr, 

A pity to see that we didn't make it into the latest Oathkeeper release 😞 

Feel free to leave additional review to push this over the finish line 💪 I think we are not far 😄

pipeline/mutate/mutator_id_token.go
```diff
@@ -208,5 +208,28 @@ func (a *MutatorIDToken) Config(config json.RawMessage) (*CredentialsIDTokenConf
 		c.TTL = "15m"
 	}
 
+	cost := int64(c.Cache.MaxCost)
+	if cost == 0 {
+		cost = 1 << 25
+	}
+
+	if a.tokenCache == nil || a.tokenCache.MaxCost() != cost {
```
David-Wobrock, line 216: The `a.tokenCache.MaxCost() != cost` condition is mainly for unit tests, in order to be able to test other `max_cost` values on the cache config.

On prod envs, this should not happen, and we'll keep skipping this branch.

pipeline/mutate/mutator_id_token.go
```diff
@@ -208,5 +208,28 @@ func (a *MutatorIDToken) Config(config json.RawMessage) (*CredentialsIDTokenConf
...
+		cache, err := ristretto.NewCache(&ristretto.Config[string, *idTokenCacheContainer]{
+			NumCounters: cost * 10,
```
aeneasr, line 218: So counters should actually be 10 * number of items. Since the cost no longer == number of items, it's not trivial to set this. But this is probably too high?

pipeline/mutate/mutator_id_token.go
aeneasr: So basically this does hot reloading if the cost changes?

.schema/config.schema.json
```diff
@@ -1140,6 +1140,25 @@
...
+              "type": "boolean",
+              "default": true,
+              "examples": [false, true],
+              "description": "En-/disables this component."
+            },
+            "max_cost": {
+              "type": "integer",
+              "default": 33554432,
+              "title": "Maximum Cached Cost",
+              "description": "Max cost to cache."
```
aeneasr, line 1159: In this case the cost is approximately the number of bytes we hold? If so maybe say that here

aeneasr (commented):
Generally LGTM, a few comments

pipeline/mutate/mutator_id_token.go
David-Wobrock: Yeah, being able to estimate the number of JWTs that we can fit in the given max size relates to the average size of JWTs. But it's gonna be tough to get an "average JWT size" when the payload is configurable 😅 

What would be a better guess in your opinion?
I pushed this for now, which would leave a bit less extra space through the number of counters.
```suggestion
			NumCounters: cost * 4,
```

aeneasr:
Looks like we're now failing some cache tests: https://github.com/ory/oathkeeper/actions/runs/12588707987/job/35087289005?pr=1177

David-Wobrock:
> Looks like we're now failing some cache tests: https://github.com/ory/oathkeeper/actions/runs/12588707987/job/35087289005?pr=1177

I think we're good again 🙂

Issue #1144, opened 2023-12-13
Decision API is not respecting the token_from config


osbornk:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/oathkeeper/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/oathkeeper/blob/master/CONTRIBUTING.md).
- [ ] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe the bug

I am trying to send a custom token header to the Oathkeeper API for token auth. But it seems to not pass a custom token header to the `check_session_url`. Instead, it does not send a token at all.

Oathkeeper does check the token. If I don't include the token, Oathkeeper immediately returns a 401. But it never gets passed onto the `check_session_url`.

### Reproducing the bug

Here is my command:
```
curl -H "X-Forwarded-Method: GET" -H "X-Forwarded-Proto: http" -H "X-Forwarded-Host: myhost" -H "X-Forwarded-Uri: /xyz" -H "xyz: sometoken" -X GET http://oathkeeper-api:4456/decisions -v
```

From my logs, my request makes it to Kratos, but without the token, so it produces a 401.

My expectation is for Oathkeeper to read the header `xxx` and send `Authorization: Bearer sometoken` to Kratos.

If I remove the `token_from` section from the config and use the default values, this works.

```
curl -H "X-Forwarded-Method: GET" -H "X-Forwarded-Proto: http" -H "X-Forwarded-Host: myhost" -H "X-Forwarded-Uri: /xyz" -H "Authorization: Bearer sometoken" -X GET http://oathkeeper-api:4456/decisions -v
```

### Relevant configuration

```yml
config:
    access_rules:
      repositories:
        - file:///etc/rules/access-rules.json
    authenticators:
      bearer_token:
        enabled: true
        config:
          check_session_url: http://kratos-public/sessions/whoami
          token_from:
            header: xxx
          extra_from: "@this"
          subject_from: "identity.traits.my_id"
    authorizers:
      allow:
        enabled: true
    mutators:
      header:
        enabled: true
        config:
          headers:
[...]

PR #1189, opened 2024-09-20
Ensure `bearer_token` Respects `token_from`

TamerShlash:
## The Problem

The `bearer_token` authenticator always returns 401 when providing any `token_from` option. The reason is because the token is indeed detected, and therefore the authenticator is deemed responsible, but the token is not actually passed forward to the session store in the `Authorization` header as the instructions [here](https://www.ory.sh/docs/identities/sign-in/check-session-token-cookie-api) and [here](https://www.ory.sh/docs/reference/api#tag/frontend/operation/exchangeSessionToken) say should be done.

The only exception when it does work is when using the default `Authorization: bearer <session-token>` option (i.e, not specifying `token_from at all), because the `Authorization` header is forwarded anyways as you can see here:

https://github.com/ory/oathkeeper/blob/acb258466eba7dd9900ea37a42e8bf98ec59a591/pipeline/authn/authenticator_bearer_token.go#L124

This PR fixes that by making sure to set `Authorization: bearer <token>` for the request going to the sessions store if a token is detected.

## Is It Breaking?

No, it should not be.

## Related issue(s)

#1144 

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/middleware/grpc_middleware_test.go
+++ b/middleware/grpc_middleware_test.go
@@ -50,7 +50,7 @@ func testClient(t *testing.T, l *bufconn.Listener, dialOpts ...grpc.DialOption) 
 func testTokenCheckServer(t *testing.T) *httptest.Server {
 	s := httptest.NewServer(http.HandlerFunc(
 		func(w http.ResponseWriter, r *http.Request) {
-			if r.Header.Get("authorization") != "Bearer correct token" {
+			if r.Header.Get("Authorization") != "bearer correct token" {
 				t.Logf("denied request %+v", r)
 				w.WriteHeader(http.StatusForbidden)
 				return
@@ -77,7 +77,7 @@ func writeTestConfig(t *testing.T, pattern string, content string) string {
 type testToken string
 
 func (t testToken) GetRequestMetadata(context.Context, ...string) (map[string]string, error) {
-	return map[string]string{"authorization": "Bearer " + string(t)}, nil
+	return map[string]string{"Authorization": "bearer " + string(t)}, nil
 }
 func (t testToken) RequireTransportSecurity() bool { return false }
 
--- a/pipeline/authn/authenticator_bearer_token.go
+++ b/pipeline/authn/authenticator_bearer_token.go
@@ -141,6 +141,11 @@ func (a *AuthenticatorBearerToken) Authenticate(r *http.Request, session *Authen
 		return errors.WithStack(ErrAuthenticatorNotResponsible)
 	}
 
+	if r.Header == nil {
+		r.Header = make(http.Header)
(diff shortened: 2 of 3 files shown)


TamerShlash:
Hello @aeneasr 👋 can you take a look at this?

zepatrik (commented):
Thanks for chiming in here! However, I'm not quite sure if this is the correct way to solve the problem.
The basic problem is that the bearer authenticator handles `token_from` in an unexpected way: the token is forwarded, but with the same method as defined in `token_from`.
Your fix adds the token to the forwarded request in the authorization header, however the original header is also forwarded. This fixes where the session store expects a `Authorization: bearer <token>`, however it does not solve other cases.
What do you think of adding a new config option to the authenticator, basically `forward_token_as` that is the inverse of `token_from`. With this approach we can ensure that we don't break anyone relying on the current behavior, but also don't forward the token twice.

I would not want to forward the token twice, as the token store might not correctly strip the token from logs etc. when it receives "just another extra header".

WDYT?

TamerShlash:
Hi @zepatrik! thank you for taking a look at this.

I think the main misunderstanding here is that in my mind a session store was equal to Ory Kratos, which expects the session token exclusively in the `Authorization` header.
What I'm understanding from you now is that this PR might break solutions where the sesison store was something other than Ory Kratos, and those solutions have been able to expect the token in other headers and thus forwarding worked fine for them.

If that's what you mean, that seems totally reasonable, and the fix you suggested with `forward_token_as` seems like a pretty reasonable fix.

My only take on this is that Oathkeeper authenticators configuration has already been quite confusing to understand to begin with (maybe just for us), so I think we need to properly document how this new `forward_token_as` option works *especially* with Ory Kratos as a session store since I'm assuming it's a major use case.

How do you want to go forward with this? I'm happy to update this PR with that suggestion if but can't tell how soon it will be since it's a bit busy for me this period and this is my first time working with Go ever 😄 Otherwise of course it's great if you want to take care of it.

mathportillo:
I'd like to weigh in by saying that my team and I also thought this should work as @TamerShlash described, and we considered the current behavior to be a bug.

Since Kratos only works with the Authorization header, token_from sounds like it would take the token from that source and send it to Kratos in the only way it supports.

However, we agree that it makes sense, for backward compatibility purposes, to add a different configuration option, forward_token_as, to fulfill this purpose.

Another option would be to add support to Kratos for reading tokens from other sources, similar to token_from. Although, this might not be ideal.

@zepatrik @aeneasr, what do you think?

We can try to help as well (albeit slowly, as we are not proficient in Go).

PR #1209, opened 2024-12-28
fix: memory leak in id_token mutator cache

David-Wobrock:
Set a TTL on the cached JWTs in the `id_token` mutator.
It fixes a memory leak in Oathkeeper 🙌 

We are running internally a fork of Oathkeeper with this patch applied, and the resulting memory footprint:
image

Unsure how to properly test it properly in a unit test though, since the `getFromCache` logic also checks if the cached TTL value is not expired (and not only the new ristretto internal TTL).

## Related issue(s)

Split from https://github.com/ory/oathkeeper/pull/1177.

## Checklist

      introduces a new feature.
- [x] I am following the
      the changes.
      works.

### DIFF
--- a/pipeline/mutate/mutator_id_token.go
+++ b/pipeline/mutate/mutator_id_token.go
@@ -112,11 +112,16 @@ func (a *MutatorIDToken) tokenToCache(config *CredentialsIDTokenConfig, session 
 	}
 
 	key := a.cacheKey(config, ttl, claims, session)
-	a.tokenCache.Set(key, &idTokenCacheContainer{
-		TTL:       ttl,
-		ExpiresAt: expiresAt,
-		Token:     token,
-	}, 0)
+	a.tokenCache.SetWithTTL(
+		key,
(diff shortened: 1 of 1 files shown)


aeneasr (approved):
Thank you! This makes sense - one remark though. If you see high memory use the cache is probably misconfigured. Ristretto does evict items when the cache size limit has been reached so _theoretically_ you don’t need TTL (but it makes sense because otherwise the cache might evict tokens that are still valid)

David-Wobrock:
> Thank you! This makes sense - one remark though. If you see high memory use the cache is probably misconfigured. Ristretto does evict items when the cache size limit has been reached so _theoretically_ you don’t need TTL (but it makes sense because otherwise the cache might evict tokens that are still valid)

Agreed 👍 
There's an attempt to tackle this here: https://github.com/ory/oathkeeper/pull/1177
Which changes the default cache configuration, but also makes the `id_token` cache configurable.

PR #1259, opened 2026-02-26
fix: restore OTEL trace propagation in remote and remote_json authorizers

rdark:
The httpx.ResilientClientWithTracer option was removed from ory/x during the monorepo migration, but the remote and remote_json authorizers were not updated to compensate. Their HTTP clients now use a plain transport with no trace context propagation, breaking distributed tracing to downstream services.

Restore propagation by wrapping the client transport with otelhttp.NewTransport after construction. This uses the global OTEL TracerProvider (already registered by otelx.SetupOTLP via otel.SetTracerProvider), consistent with how authenticator_cookie_session.go handles the same requirement.

The fix applies to both the initial client in the constructor and the retry client rebuilt in Config() when timeout/max-wait settings are present.

NOTE: Two stale entries in go.mod pruned by go mod tidy

## Checklist

      introduces a new feature.
- [x] I am following the
      [security@ory.com](mailto:security@ory.com)) from the maintainers to push
      the changes.
      works.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,6 +63,7 @@ jobs:
     needs:
       - test
     steps:
+      - uses: ory/ci/checkout@master
       - uses: ory/ci/docs/cli-next@master
         with:
           token: ${{ secrets.ORY_BOT_PAT }}


aeneasr:
Thank you! We still have some work to do to get the monorepo set up working 100% for the OSS pipeline, sorry about that. Also @hperl not sure how that commit was dropped, did we forget to import some git hash?

